PR #7, opened 2016-09-27
Add Dependencies Badge via Gemnasium

athix:
This probably shouldn't be merged until we figure out how we want to support both Rails 4 and Rails 5 at the same time. (the out-of-date warning is because we use Rails ~> 4)


Ch4s3:
Yeah, that sounds reasonable

PR #10, opened 2016-10-03
Slack auth

youzik:
Just write a new provider to slack oauth2.  tests are green.

### DIFF
--- a/sorcery.gemspec
+++ b/sorcery.gemspec
@@ -1,63 +1,36 @@
-# -*- encoding: utf-8 -*-
-# stub: sorcery 0.9.1 ruby lib
+lib = File.expand_path('../lib', __FILE__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require 'sorcery/version'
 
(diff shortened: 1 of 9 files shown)


athix:
Hey @youzik, thanks for the contribution! I'll try to take a look at this when I have the time. If you're interested, we also need to update our oauth for microsoft to use their new API. (Live has been deprecated in favor of liveconnect iirc)

sorcery.gemspec
```diff
@@ -1,37 +1,63 @@
-lib = File.expand_path('../lib', __FILE__)
```
arnvald: @youzik please revert all the changes in this file, they're not needed and some of them break the current way the gem is configured (e.g. it hardcodes the version of the gem instead of loading it from `Sorcery::VERSION`)

arnvald:
Except for the changes in `gemspec` it looks good to me, but I haven't tested it live

youzik:
@arnvald ok, gemspec changed.

Ch4s3:
Looks good.

PR #13, opened 2016-10-25
delete reset_password_email_sent_at when change_password

syossan27:
I fixed that reset_password_email_sent_at isn't nil when the change_password!.
because, can't sent a mail at the time of the second time.


kyuden:
@syossan27  `change_password!` doesn't send mail, so It is strange that `reset_password_email_sent_at` becomes `nil`. Please check how to use `change_password!` In this [tutorial](https://github.com/Sorcery/sorcery/wiki/Reset-password).

syossan27:
@kyuden Thanks reply!
Sorry... My recognition was wrong.
close this pull request.

kyuden:
@syossan27  Thank you for your contribution!

PR #19, opened 2016-11-14
Use urlsafe_base64 for url sensitive chars

kyuden:
We can use the urlsafe one.

### DIFF
--- a/lib/sorcery/model/temporary_token.rb
+++ b/lib/sorcery/model/temporary_token.rb
@@ -12,7 +12,7 @@ module Sorcery
 
       # Random code, used for salt and temporary tokens.
       def self.generate_random_token
-        SecureRandom.base64(15).tr('+/=lIO0', 'pqrsxyz')
+        SecureRandom.urlsafe_base64(15).tr('lIO0', 'sxyz')
       end
 
       module ClassMethods


lib/sorcery/model/temporary_token.rb
```diff
@@ -12,7 +12,7 @@ def self.included(base)
 
       # Random code, used for salt and temporary tokens.
       def self.generate_random_token
-        SecureRandom.base64(15).tr('+/=lIO0', 'pqrsxyz')
+        SecureRandom.urlsafe_base64(15).tr('lIO0', 'sxyz')
```
athix: Using urlsafe here is definitely preferable, although I wonder if we can deprecate this model entirely by using rails built in `has_secure_token`?

kyuden:
The behavior is the almost same around this change.
Are we supposed to need an entry on CHANGELOG.md in this case?

lib/sorcery/model/temporary_token.rb
kyuden: When we drop rails 4.x supports, we might be able to deprecate it with `has_secure_token`.

athix:
Unless I'm mistaken, this changes behavior that could affect how someone handles the tokens (if it's not URL safe to begin with, someone might be running a strip method before using it in URLs). It likely won't affect most users, but I would err on the side of caution.

Issue #20, opened 2016-11-21
Rails 5 Release


athix:
Creating this issue to discuss releasing a version that fixes all the rails 5 deprecation warnings. Probably won't have time to develop some of the more major items, so I was thinking maybe call this release 0.10.0 and have it be the last release before we go through the existing code with a fine tooth comb and modularize the various features (1.0.0). IIRC, master branch solves most if not all of the problems, so it will be more going through and verifying it's production ready.

@Ch4s3, what are your thoughts and availability? I'll likely be able to dedicate large amounts of time over the next week or two to getting this done.

athix:
Running into this issue when trying to use the latest version in my app, any ideas?

<img width="508" alt="screen shot 2016-12-22 at 2 45 30 pm" src="https://cloud.githubusercontent.com/assets/1802598/21443002/8a0a477c-c856-11e6-8b94-04511b3149f1.png">

kyuden:
@athix  Did you try to do `bundle update`?

dankimio:
Should we weaken dependency requirements?

athix:
@kyuden This error is from running `bundle update`.

I've looked into this a bit more, and it's looking like oauth2's outdated dependencies are the cause of the issue. There's open pull requests to fix them (`intridea/oauth2` [#277](https://github.com/intridea/oauth2/pull/277),[#278](https://github.com/intridea/oauth2/pull/278)), but until that happens I don't really know what we can do to fix it.

athix:
Resolved the issue by locking the twitter gem to `~> 5.0`.

PR #21, opened 2016-11-27
Test against Rails 5 by default

kyuden:
This PR pass all tests and fix deprecation warnings on Rails 5.

In the process, I removed test of rails 4.0.1. https://github.com/Sorcery/sorcery/commit/b69a89bbe0e6ed2aac6a9fc3694ce1bfaf248a71
Because `render plain` did not exist on rails 4.0.1, but It needs on Rails 5. https://github.com/Sorcery/sorcery/commit/4fc164f2e285ca1f85d16497113b8730bedd3041
I can also fix this problem with the way like [this](https://github.com/Sorcery/sorcery/compare/master...kyuden:rails5?expand=1#diff-d0d1d0d5ee17b0aee49ab5b831d7e9a3R66) , but rails 4.0.1 have not been maintained yet. So I removed test of rails 4.0.1.  What do you think?

related: https://github.com/Sorcery/sorcery/issues/20

edit:
I noticed test of rails 4.0.1 don't need to be drop. So, I added a test of rails 5 only.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,12 @@ matrix:
     - rvm: 2.0.0
       gemfile: gemfiles/active_record-rails42.gemfile
 
+    - rvm: 2.0.0
+      gemfile: Gemfile
+
+    - rvm: 2.1
+      gemfile: Gemfile
+
     - rvm: jruby
       gemfile: Gemfile
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,9 @@ language: ruby
 rvm:
   - jruby
   - 2.0.0
-  - 2.1
-  - 2.2
-  - 2.3.0
+  - 2.1.10
+  - 2.2.6
+  - 2.3.3
 
 env:
   global:
@@ -34,14 +34,14 @@ matrix:
     - rvm: 2.0.0
       gemfile: Gemfile
 
-    - rvm: 2.1
+    - rvm: 2.1.10
       gemfile: Gemfile
+
+    - rvm: 2.2.6
+      gemfile: gemfiles/active_record-rails40.gemfile
+
+    - rvm: 2.3.3
+      gemfile: gemfiles/active_record-rails40.gemfile
 
     - rvm: jruby
       gemfile: Gemfile
-
-    - rvm: 2.2
-      gemfile: gemfiles/active_record-rails40.gemfile
-
-    - rvm: 2.3.0
-      gemfile: gemfiles/active_record-rails40.gemfile
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'rails', '~> 4.0'
+gem 'rails', '~> 5.0.0'
 gem 'sqlite3'
 gem 'pry'
 
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,7 @@
 source 'https://rubygems.org'
 
 gem 'rails', '~> 5.0.0'
+gem 'rails-controller-testing'
 gem 'sqlite3'
 gem 'pry'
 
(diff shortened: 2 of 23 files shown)


athix:
Thanks! My time is limited, but I'll try to review then merge this when I get a chance.

kyuden:
@athix  I'm looking forward to it :) 
There are many PRs not reviewed. If you add me to the sorcery organization, I will do it in my free time. I think the sorcery organization needs more maintainers.

athix:
@kyuden, I definitely agree. I'll add you to the org, I'd just ask that you refrain from merging PRs without discussing it with @Ch4s3 and/or myself first.

kyuden:
@athix Thanks!, That would be fine.

.travis.yml
```diff
@@ -1,18 +1,16 @@
 language: ruby
 rvm:
   - jruby
-  - 2.0.0
-  - 2.1
-  - 2.2
-  - 2.3.0
+  - 2.1.10
+  - 2.2.6
+  - 2.3.3
```
kyuden: > we should also keep support for 2.0

Okay. I changed to support it

Gemfile
```diff
@@ -1,6 +1,7 @@
 source 'https://rubygems.org'
 
-gem 'rails', '~> 4.0'
+gem 'rails', '~> 5.0.0'
+gem 'rails-controller-testing'
```
kyuden: > We should try to minimize dependencies wherever possible.

I agree. But, this gem is very useful for passing controller spec on both Rails4.x and Rails5. Besides, this gem is maintained by rails organization. I think that it is better to use it.

sorcery.gemspec
```diff
@@ -30,7 +30,6 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency "timecop"
   s.add_development_dependency "simplecov", ">= 0.3.8"
-  s.add_development_dependency "rspec", "~> 3.1.0"
-  s.add_development_dependency "rspec-rails", "~> 3.1.0"
+  s.add_development_dependency "rspec-rails", "~> 3.5"
```
kyuden: I fixed it.

.travis.yml
kyuden: @athix Many other gems also have used this approach.

athix: @kyuden, I'm sure what you have is fine. I just want to verify that we weren't using 2.1 and 2.2 for a specific reason and won't have time until this weekend to do so.

Thanks again for getting this in, the only blocker I see right now is the travis versioning; so it should be merged soon.

kyuden: Thanks!

Gemfile
dankimio: I think we should keep the dependency. Since we're upgrading to Rails 5, it will make tests passing on both Rails 4 and 5, otherwise we'd have to manually update all controller tests for Rails 5. The gem itself is not too heavy.

.travis.yml
kyuden: @athix  Are you busy this week?

athix: From what I can tell, using 2.x should use the latest patch level version of x. Can we try using `2.0`, `2.1`,`2.2`, and `2.3` to see how it handles each? If any of them fail to use the latest version, I'd say it would be fine to use the patch specification, otherwise it's easier to maintain if we keep it to minor only.

athix: Ah, sorry forgot about that. I'll play around with travis on my fork and see if I can get the fuzzy version locking playing nicely. For now, I'll go ahead and merge this in.

dankimio: I think we should use the latest patch version possible. I don't see any point in testing on the older versions. Also, Ruby 2.0.0 is not supported anymore, so it's a good idea to specify the latest patch version. This PR should be merged, Ruby versions for `.travis.yml` can be discussed separately.

kyuden:
@athix  Thank you for your reply in your busy time :)

PR #17, opened 2016-11-08
Refactor by using rubocop

asmsuechan:
Before
```
113 files inspected, 2864 offenses detected
```

After
```
113 files inspected, 2768 offenses detected
```

We should discuss about .rubocop.yml.

PR #23, opened 2016-12-12
Code style

dankimio:
Fixes another ~2000 warnings in addition to #17.


kyuden:
@dankimio  We need to think about the rubocop configuration. and then, we should fix the codes with the rubocop configuration. Do you want me to create a sample of configuration? If it's is the case, I’ll close this PR and create new one.

dankimio:
@kyuden There's no need to. I've used the default one provided by Rubocop. Now that almost all  warnings are fixed, it is ready to be merged. The only ones left are regarding long lines and top-level comments, which should be okay for now.

kyuden:
@dankimio  You should also use rubocop on CI for continuous maintenance. So, you have to push `.rubocop.yml` to root.  But default one is a little noisy. What do you think?

dankimio:
@kyuden I agree with you on CI, but I wanted to focus this PR on fixing the existing style issues, so there's no need for a config file right now.

I'm pretty happy with the Rubocop defaults (with a few exceptions). It would be great to know what @athix @Ch4s3 think about adopting some style guide other than the [default](https://github.com/bbatsov/ruby-style-guide) one (or customizing).

kyuden:
@dankimio  I recommend not the default one but the rails [one](https://github.com/rails/rails/blob/master/.rubocop.yml#L3-L5).

PR #27, opened 2016-12-20
Add Wechat provider

Darmody:
Add Wechat provider. 😄 

Resolved #27

### DIFF
--- a/lib/generators/sorcery/templates/initializer.rb
+++ b/lib/generators/sorcery/templates/initializer.rb
@@ -121,6 +121,10 @@ Rails.application.config.sorcery.configure do |config|
   # config.paypal.callback_url = "http://0.0.0.0:3000/oauth/callback?provider=paypal"
   # config.paypal.user_info_mapping = {:email => "email"}
   #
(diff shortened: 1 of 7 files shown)


lib/generators/sorcery/templates/initializer.rb
```diff
@@ -121,6 +121,10 @@
   # config.paypal.callback_url = "http://0.0.0.0:3000/oauth/callback?provider=paypal"
   # config.paypal.user_info_mapping = {:email => "email"}
   #
+  # config.wechat.key = "
```
dankimio: Can you add the closing quotation marks? So that it works after uncommenting.

Darmody: Sure, updated.

Ch4s3:
This seems fine if @dankimio and @athix are ok with it. I know we're talking about moving providers to separate gems, but we haven't moved forward on that, and wechat is very popular so a lot of folks might find it useful.

Darmody:
@athix I have test it with my own app, it works!

And CHANGELOG.md updated.

PR #41, opened 2017-01-22
Resurrect block authentication

Fustrate:
This was never merged into the old repo, but I still think it's a worthwhile change.

https://github.com/NoamB/sorcery/pull/704

These changes allow a website to tell the difference between a wrong password, a locked account, and an inactive account, so that the user can be sent to a "resend activation" or "unlock account" page when necessary.

### DIFF
--- a/lib/sorcery/controller.rb
+++ b/lib/sorcery/controller.rb
@@ -30,8 +30,16 @@ module Sorcery
       # Runs hooks after login or failed login.
       def login(*credentials)
         @current_user = nil
-        user = user_class.authenticate(*credentials)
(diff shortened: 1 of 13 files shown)


Ch4s3:
Ok, I'll take a look. Probably good if @athix signs off as well.

athix:
I'll take a look when I get a chance, definitely sounds like a good change to have though.

Fustrate:
Any updates on this?

Ch4s3:
@Fustrate sorry for the delay, this looks super helpful upon further review!  Thanks!

PR #45, opened 2017-02-06
Re-add MongoID Adapter

### LINKED ISSUE #35: Error: Undefined method `authenticates_with_sorcery!' for User:Class
I get the "Undefined method `authenticates_with_sorcery!' for User:Class" error in Rails 5.0.1 with Mongoid 6.0.3 and Sorcery 0.10.0. I added the authenticates_with_sorcery! beneath the includes in my model but still getting the error. As long as I use Rails 5.0.1, Mongoid 6.0.3 and Sorcery 0.9.1 it works fine. Is there any change I have to make with loading the functions?

Regards,
Sven

athix:
There were plans on moving adapters out, but until that actually happens we need to keep this in otherwise MongoID will break for users. Interestingly, the additional adapters were removed in 2014, but were still a part of the v0.9.1 release.

Note to users experiencing issues: After updating to Sorcery v0.10.3, you must `include Mongoid::Document` _before_ you call `authenticates_with_sorcery!`

I have not thoroughly tested this, any feedback is appreciated.

Fixes #35

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.10.3
+
+* Revert removal of MongoID Adapter (breaks Sorcery for MongoID users until separate gem is created)
+
 ## 0.10.2
 
 * Added support for Microsoft OAuth (thanks to @athix) [#37](https://github.com/Sorcery/sorcery/pull/37)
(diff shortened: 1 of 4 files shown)


athix:
@Ch4s3, when you get a chance would you mind reviewing this for release?

Thanks!

PR #49, opened 2017-02-16
fixed: 401 error in Twitter Login 

funaota:
As I tried "Twitter Login" as described in the following tutorial,
There was a 401 error with `if @ user = login_from (provider)`.

tutorial
https://github.com/Sorcery/sorcery/wiki/External

As I follow the cause,
The value of session in `Sorcery :: Providers :: Twitter.process_callback` was nil.

So, when I changed the value management to instance variable instead of session, I succeeded in "twitter login".

I tried the version of rails with 5.0.0 and 4.2.0, but the same bug(401 error) occurred in both cases.

### DIFF
--- a/lib/sorcery/providers/twitter.rb
+++ b/lib/sorcery/providers/twitter.rb
@@ -16,6 +16,8 @@ module Sorcery
 
         @site           = 'https://api.twitter.com'
         @user_info_path = '/1.1/account/verify_credentials.json'
+        @request_token = nil
+        @request_token_secret = nil
       end
 
       # Override included get_consumer method to provide authorize_path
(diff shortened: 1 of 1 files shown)


athix:
Hi @funaota, is your development machine's clock synced and callback URL on twitter set to `127.0.0.1:3000` or `0.0.0.0:3000`? (Alternatively, set the callback url to a real domain and allow applications to overwrite the callback, then in your application use one of the two local IPs) While I made a couple of adjustments afterwards, the tutorial didn't cause any issues for me so I'm curious how to replicate the issue.

funaota:
thanks @athix !
my development machine's clock synced.
and, Callback url is set to 0.0.0.0: 3000.

However, an error has occurred.

By sending a request twice with the same request token,
I think that "401 Authorization Required" has occurred.

________________________________

```
# /lib/sorcery/providers/twitter.rb

# calculates and returns the url to which the user should be redirected,
# to get authenticated at the external provider's site.
def login_url(_params, session)
  req_token = get_request_token
  session[:request_token]         = req_token.token
  session[:request_token_secret]  = req_token.secret
  authorize_url(request_token: req_token.token, request_token_secret: req_token.secret)
end
``` 
It seems that the value of the request token is assigned to `session`.

```
# /lib/sorcery/providers/twitter.rb

      # tries to login the user from access token
      def process_callback(params, session)
        args = {
          oauth_verifier:       params[:oauth_verifier],
          request_token:        session[:request_token],
          request_token_secret: session[:request_token_secret]
        }

        args[:code] = params[:code] if params[:code]
        get_access_token(args)
      end
```
Acquiring an access token In this method,
The value of session is empty.

funaota:
I am sorry that I did not have much knowledge about thread safe and made a "pull request" with a wrong solution.
I noticed that using instance variables is not good.

ebihara99999:
I don't know about the source @funaota  modified, but I replicate the same 401 error with the code which seems to have worked once. I'll looked into the source soon.

ebihara99999:
@funaota, I met the same error you encounterd (the point that raises error seems same: session is nil), but I found a solution. If you write  `config.twitter.callback_url = "http://0.0.0.0:3000/oauth/callback?provider=twitter"'`
then, you need to access `http://0.0.0.0:3000`. I encoutered the error when I set the value, but I accessed to `http://localhost:3000`. Session is not stored between different domain.

funaota:
@ebihara99999 
Thanks for telling me the solution!
It worked normally with its solution.

Issue #48, opened 2017-02-16
Is it fine to add undocumented option to initializer.rb?


tackeyy:
As I found the undocumented option which is about Github's [scope](https://developer.github.com/v3/oauth/#scopes) with external module in Github OAuth. I think it is better if below code is inserted in  [initializer.rb](https://github.com/Sorcery/sorcery/blob/master/lib/generators/sorcery/templates/initializer.rb#L114), because it takes time to find the option out if someone who uses Github OAuth with external module and want to manage the Github's scope.

> config.github.scope = 'user, repo'

athix:
I don't see any problem adding the scope to the generators, personally. Adding a note above the Github values though saying why/how to use the scope would help users understand why it's there (and why or why not they should use it). Feel free to open a PR, where we can discuss the finer details.

Also, if you wanted to update the docs in the wiki at all your help would be greatly appreciated. :)

tackeyy:
Thank you for the reply. As I opened PR #50, I close this issue.

PR #50, opened 2017-02-16
Add github scope option to initializer.rb

tackeyy:
As I found the undocumented option which is about Github's scope with external module in Github OAuth. I think it is better if this pull request is accept, because it takes time to find the option out if someone who uses Github OAuth with external module and want to manage the Github's scope.
I don't want to change original behavior when ` comment out ` of added option is remove, so default value is blank. (blank scope is called [ ` no scope ` ](https://developer.github.com/v3/oauth/#scopes).)

refs: #48

### DIFF
--- a/lib/generators/sorcery/templates/initializer.rb
+++ b/lib/generators/sorcery/templates/initializer.rb
@@ -115,6 +115,7 @@ Rails.application.config.sorcery.configure do |config|
   # config.github.secret = ""
   # config.github.callback_url = "http://0.0.0.0:3000/oauth/callback?provider=github"
   # config.github.user_info_mapping = {:email => "name"}
+  # config.github.scope = ""
   #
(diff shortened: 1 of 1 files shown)


athix:
Seeing as this doesn't change the default value, I don't really see any problem adding this even without added documentation in the config file. I feel like setting up applications for all the different oauth providers isn't very well documented, so that's something that we should probably add to the wiki. (Links to the appropriate dev docs (scopes, available mappings, etc) and simple walkthrough for getting your app started on each provider.)

tackeyy:
@athix thanks for approve.

> I feel like setting up applications for all the different oauth providers isn't very well documented, so that's something that we should probably add to the wiki.

I agree. It's more useful if all (or popular) oauth providers is well documented. if I could updated github oauth providers document, I'll open other PR :)

PR #51, opened 2017-02-18
Add Instagram Provider

Waynenabors:
My first pull request so hopefully it makes sense. 
This was started on the old repo here: https://github.com/NoamB/sorcery/pull/625/files

We are using this code with an approved Instagram client_id with public_scope and it works great.


Ch4s3:
@Waynenabors Sorry for the delay! @athix What are your thoughts on rolling this our vs making it a separate gem?

Waynenabors:
Fixes an issue where new Facebook Apps with API v2.8 are no longer compatible with v2.2 Oauth callback content/type. Starting with v2.3 the response is valid JSON. So if we can pick our API version we should be able to pick our parser. I would even make the argument that it should default to :automatic since that works now. But for now I just needed access to set `config.facebook.parse`.

mzp:
> Fixes an issue where new Facebook Apps with API v2.8 are no longer compatible with v2.2 Oauth callback content/type.

I met this problem (and monkey patched). I appreciate if this facebook's changes are merged.

Ch4s3:
If someone can resolve the conflict, I'll merge this.

athix:
Last update almost year ago, whoops. 😬

PR #54, opened 2017-03-29
Add build_from method

athix:
Currently there's no way to build a user from an oauth callback, you have to either manually map the data or create the user with create_from.


athix:
I think this should be ready for merging, @Ch4s3 mind giving this a quick once over when you have a minute?

Doesn't have any specs yet, but it shouldn't affect any existing functionality so that can probably wait for when we redo the overall specs.

PR #55, opened 2017-04-01
Change default value of config.github.user_info_mapping

asmsuechan:
I'd like to change the default value of `config.github.user_info_mapping` because the value of name is duplicatable on GitHub and the default key (`email`) cannot duplicate. What do you think about this?

### DIFF
--- a/lib/generators/sorcery/templates/initializer.rb
+++ b/lib/generators/sorcery/templates/initializer.rb
@@ -114,7 +114,7 @@ Rails.application.config.sorcery.configure do |config|
   # config.github.key = ""
   # config.github.secret = ""
   # config.github.callback_url = "http://0.0.0.0:3000/oauth/callback?provider=github"
-  # config.github.user_info_mapping = {:email => "name"}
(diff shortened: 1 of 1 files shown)


lib/generators/sorcery/templates/initializer.rb
```diff
@@ -114,7 +114,7 @@
   # config.github.key = ""
   # config.github.secret = ""
   # config.github.callback_url = "http://0.0.0.0:3000/oauth/callback?provider=github"
-  # config.github.user_info_mapping = {:email => "name"}
+  # config.github.user_info_mapping = {:email => "login"}
```
athix: Are we sure that `login` provides the user's contact email, and doesn't require additional an permission scope? I know that github treats its login email separately from the public profile email, so might be worth looking into further.

athix:
I tried looking into this myself, but the Github documentation is so all over the place when it comes to OAuth that I can't find what their return values are. Closing this PR for now unless someone shows renewed interest.

Issue #40, opened 2017-01-20
Password reset token leak via HTTP referer


williantenfen:
The reset password token is leaking through the HTTP referer header ... This happens when user clicks at the link sent to their email and when the page is rendered with the token at the URL, if there is some subsequent requests to CDN/analytics or malicious JS  , the token will be present at the REFERER header of this requests... As reported at here: [https://robots.thoughtbot.com/is-your-site-leaking-password-reset-links](https://robots.thoughtbot.com/is-your-site-leaking-password-reset-links)  A solution will be to store the token to session and then redirect to the page without the token at the URL ... 
Is there any chance to change this behavior to current versions?

athix:
If you open a PR for this, we can definitely take a look and push out a new version. We'd also want to push to the 0.9.x branch and fix the security issue if possible.

ebihara99999:
Hi @athix 
I looked over the issue, wondering which way is favorable, and I would be happy if you offer a suggestion.

The link offers 2 solutions: session based approach and immediate invalidation approach.
The first one is, as @williantenfen mentioned, is to store a token in session. This needs to modify documentation or templates because the password reset feature is implemented only in the model layer. 
To embed a token to session instead of url, the following needs to be done:  
* Modify `resources :password_resets` to `resource :password_resets` in `config/routes.rb`.
* Modify `edit_password_reset_url(@user.reset_password_token)` to `edit_password_reset_url` in `app/mailers/user_mailer.rb` 
* Add a token to session after `@user.deliver_reset_password_instructions!` succeeds in `app/controllers/password_resets_controller.rb`.
* Modify `params[:id]`, which stores a token, to session[:token] not to depend on url parameters including a token in `app/controllers/password_resets_controller.rb`.

To complete these instructions, it's required to modify documentation and generated templates.

The second approach is achieved by adding visited flag to database or session and modify `#load_from_reset_password_token` to set the flag. At the same time, it's necessary to remove implementions related to `reset_password_token_expires_at` and `reset_password_email_sent_at` because this way no longer needs them.

ebihara99999:
I'm sorry I don't intend to reference the commits from this issue. It's a mistake.

ebihara99999:
Hi @Ch4s3, would you suggest the way to go? I prefer the second way; because the most of the fix in the first way is about the documentation, it seems difficult to check if the modification is correct.

ebihara99999:
Hi @athix and @Ch4s3
[...]

PR #56, opened 2017-04-08
Fix issue #40

ebihara99999:
Fix [Password reset token leak via HTTP referer](https://github.com/Sorcery/sorcery/issues/40).
This change requires to update [the documentation](https://github.com/Sorcery/sorcery/wiki/Reset-password) accordingly, and I will explain how to fix it in implementation and the documentation (written below).

## Implementation
I follow the step of the first direction mentioned in [Is Your Site Leaking Password Reset Links?](https://robots.thoughtbot.com/is-your-site-leaking-password-reset-links#testing-your-site): 
>Update the passwords#edit controller action so it immediately invalidates the password reset token in the request and generates a new one that is used in the form action. The Referer will still contain the original password reset token, but the token is no longer valid.

* Add access_count_to_reset_password_page attribute to user model, responsible for the count how many times the password reset page is accessed.
In `lib/sorcery/model/submodules/reset_password.rb`, 
* Add `#increment_password_reset_page_access_counter`, which increments access_count_to_reset_password_page attribute.
* Add `#reset_password_reset_page_access_counter`. This resets access_count_to_reset_password_page attribute to 0

I will explain how these changes affect the password reset feature in a rails application below.

## Documentation
The wiki is expected to update as follows only `PasswordResetsController#create` and `PasswordResetsController#edit`:
```
class PasswordResetsController < ApplicationController
  skip_before_filter :require_login
  
  def create
    @user = User.find_by_email(params[:email])
    
    #### Difference Start#####
    # Sends an email to the user with instructions on how to reset their password (a url with a random token)
    # Don't forget to reset password_reset_page_access_counter
    # For security reset password page can be accessed only once.
    if @user
      @user.deliver_reset_password_instructions!
      @user.reset_password_reset_page_access_counter
    end
    #### Difference End #####
    
    # Tell the user instructions have been sent whether or not email was found.
    # This is to not leak information to attackers about which emails exist in the system.
    redirect_to(root_path, :notice => 'Instructions have been sent to your email.')
  end
  
  # This is the reset password form.
  def edit
    @token = params[:id]
    @user = User.load_from_reset_password_token(params[:id])
    
    if @user.blank?
      not_authenticated
      return
    end
    
    #### Difference Start #####
    # Increment password reset page access counter
    @user.increment_password_reset_page_access_counter

    #  password reset page can't be accessed more than twice
    if @user.access_count_to_reset_password_page > 1
      # It's nice to override #not_authenticated to show a message of the reason not authenticated.
      not_authenticated
      return
    end
    #### Difference End #####
  end
end
```

### Sorcery
All specs are green.
### Documentation
Test the steps mentioned in [Is Your Site Leaking Password Reset Links?](https://robots.thoughtbot.com/is-your-site-leaking-password-reset-links#testing-your-site), confirming the second access is invalid.
>1. Request a password reset and click the link that is emailed to you.
>2. Copy the URL from the resulting page.
>3. Open a private browser window or a different web browser and paste the copied URL.
>If a working password reset form is rendered then you have proven that without any mitigating steps, any Referer generated by this page would contain a working password reset URL.

### DIFF
--- a/lib/generators/sorcery/templates/initializer.rb
+++ b/lib/generators/sorcery/templates/initializer.rb
@@ -357,6 +357,11 @@ Rails.application.config.sorcery.configure do |config|
     # Default: `5 * 60`
     #
     # user.reset_password_time_between_emails =
+    
+    # access counter to a reset password page attribute name
+    # Default: `:access_count_to_reset_password_page`
+    #
+    # user.reset_password_page_access_count_attribute_name =
 
     # -- brute_force_protection --
     # Failed logins attribute name.
--- a/lib/generators/sorcery/templates/migration/reset_password.rb
+++ b/lib/generators/sorcery/templates/migration/reset_password.rb
@@ -3,6 +3,7 @@ class SorceryResetPassword < <%= migration_class_name %>
     add_column :<%= model_class_name.tableize %>, :reset_password_token, :string, :default => nil
     add_column :<%= model_class_name.tableize %>, :reset_password_token_expires_at, :datetime, :default => nil
     add_column :<%= model_class_name.tableize %>, :reset_password_email_sent_at, :datetime, :default => nil
+    add_column :<%= model_class_name.tableize %>, :access_count_to_reset_password_page, :integer, :default => 0
 
     add_index :<%= model_class_name.tableize %>, :reset_password_token
   end
--- a/lib/sorcery/model/submodules/reset_password.rb
+++ b/lib/sorcery/model/submodules/reset_password.rb
@@ -109,8 +109,8 @@ module Sorcery
           end
           
           # Increment access_count_to_reset_password_page attribute.
-          # For example, access_count_to_reset_password_page attribute is over 1, which means
-          # some one doesn't have a right to access.
+          # For example, access_count_to_reset_password_page attribute is over 1, which
+          # means the user doesn't have a right to access.
           def increment_password_reset_page_access_counter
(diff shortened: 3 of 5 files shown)


ebihara99999:
@athix 
I know you have a lot to do, thanks for letting me notice.
This is a security issue (though the risk is low) and needs to be released.
It's been long, I would like to explain the point.

The risk:
- the password reset url including a reset token, is set to refferer when a user accesses to the passoword reset page
- the url is leaked if the user accesses another link from the password reset page: considered the case of the access to external links from the footer

Considering these points, barring the leak is achieved by the restriction of one time access. 
Because [the report](https://robots.thoughtbot.com/is-your-site-leaking-password-reset-links) offers 2 solutions - the restriction of one time access and including a token in session - let me compare them.

The restriction of one time access is:
- safe because the token is invalidated even though it's leaked
- more simple, more understandable and easier to be tested

I think it is the better way to prevent the leak, and I would appreciate if you move on to reviewing my codes.

ebihara99999:
@athix 
And I can help you more with dealing with and closing issues, reviewing other PRs if you let me in a member. I would appreciate if you think about it.

athix:
@Ch4s3 do you have the time to review/merge this? I probably won't be able to get to it for at least a couple weeks.

Ch4s3:
@athix and @ebihara99999 this looks good. I'll merge it once I have some time to update the docs.

Ch4s3:
I'll try to merge this tonight

PR #57, opened 2017-04-26
Add the request spec helper method, #login_user.

ebihara99999:
There is no test helper for request specs now. It' not convenient officially recommending to write request specs to a new rails app.
If this PR is merged, I will add the documentation how to use the helper.

Thank you for review.

### DIFF
--- a/lib/sorcery/test_helpers/rails/request.rb
+++ b/lib/sorcery/test_helpers/rails/request.rb
@@ -1,0 +1,20 @@
+module Sorcery
+  module TestHelpers
+    module Rails
+      module Request
+        # Accepts arguments for user to login, the password, route to use and HTTP method
+        # Defaults - @user, 'secret', 'user_sessions_url' and http_method: POST
+        def login_user(user = nil, password = 'secret', route = nil, http_method = :post)
+          user ||= @user
+          route ||= user_sessions_url
+
+          username_attr = user.sorcery_config.username_attribute_names.first
+          username = user.send(username_attr)
+          password_attr = user.sorcery_config.password_attribute_name
+
+          self.send(http_method, route, params: { "#{username_attr}": username, "#{password_attr}": password })
+        end
+      end
+    end
+  end
+end
--- a/lib/sorcery/test_helpers/rails/request.rb
+++ b/lib/sorcery/test_helpers/rails/request.rb
@@ -12,7 +12,7 @@ module Sorcery
           username = user.send(username_attr)
           password_attr = user.sorcery_config.password_attribute_name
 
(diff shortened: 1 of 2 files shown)


ebihara99999:
checks are failed...should I use the hash rocket?

athix:
@ebihara99999 try merging/rebasing the upstream changes and see if that fixes travis. (Now that we've removed the older version on HEAD)

ebihara99999:
@athix 
Codacy failed because of using Ruby 2.1 parser. Would you tell me how to configure codacy not using ruby 2.1?

athix:
@ebihara99999 Sorry for the late reply.

I can't see an easy way to configure the ruby version from codacy, so I'm adding a rubocop config that should hopefully solve the issue. From there we should be able to rebase this branch and merge.

ebihara99999:
@athix Thanks for dealing with it!

ebihara99999:
At first I rebased the upstream changes, but it kept failing. But I found it was able to be fixed by removing the `self`, which raises any errors. The cause of the codacy failure is redundant `self`, I should've come up with how to fix it. I appreciate https://github.com/Sorcery/sorcery/pull/107 when you are busy. Anyway, it seems ready to be merged. Would you work on merge?

lib/sorcery/test_helpers/rails/request.rb
```diff
@@ -0,0 +1,20 @@
+module Sorcery
+  module TestHelpers
+    module Rails
+      module Request
+        # Accepts arguments for user to login, the password, route to use and HTTP method
+        # Defaults - @user, 'secret', 'user_sessions_url' and http_method: POST
+        def login_user(user = nil, password = 'secret', route = nil, http_method = :post)
```
athix: I wonder if secret being the password default makes the most sense here...I can't think of anything that would make more sense off the top of my head though, so I think it should be fine for now.

ebihara99999: I agree with the password default value. I added modification from Integration#login, which forces us to use ‘secret’. I don’t think the deafault value isn’t needed because what is really needed is to choose the password value, and I will remove the value if necessary.

ebihara99999:
I also have the same opinion for testing. I'm sorry I don't add the test.

But even more, IMO, I would like to give some modifications to tests now step by step if I have time. Sometimes testing is not easy because test codes are a little bit old.

Ch4s3:
This looks good to me.

ebihara99999:
Thank you for saying that, even though I'm just a contributor and can't close issues and merge PRs, I feel free to move thing forward @Ch4s3

anthony-robin:
Any news on this ?

PR #67, opened 2017-06-13
Configurable token randomness

mtsmfm:
To increase security, I want to change token randomness by configuration 🔑

### DIFF
--- a/lib/generators/sorcery/templates/initializer.rb
+++ b/lib/generators/sorcery/templates/initializer.rb
@@ -29,6 +29,12 @@ Rails.application.config.sorcery.configure do |config|
   #
   # config.remember_me_httponly =
 
+  # Set token randomness. (e.g. user activation tokens)
+  # The length of the result string is about 4/3 of `token_randomness`.
+  # Default: `15`
+  #
+  # config.token_randomness =
+
   # -- session timeout --
   # How long in seconds to keep the session alive.
   # Default: `3600`
@@ -249,12 +255,6 @@ Rails.application.config.sorcery.configure do |config|
     # Default: `false`
     #
     # user.subclasses_inherit_config =
-
-    # Set token randomness.
-    # The length of the result string is about 4/3 of `token_randomness`.
-    # Default: `15`
-    #
-    # user.token_randomness =
 
     # -- remember_me --
     # How long in seconds the session length will be
--- a/lib/generators/sorcery/templates/initializer.rb
+++ b/lib/generators/sorcery/templates/initializer.rb
@@ -250,6 +250,12 @@ Rails.application.config.sorcery.configure do |config|
     #
     # user.subclasses_inherit_config =
 
+    # Set token randomness.
+    # The length of the result string is about 4/3 of `token_randomness`.
+    # Default: `15`
+    #
+    # user.token_randomness =
+
     # -- remember_me --
     # How long in seconds the session length will be
     # Default: `604800`
--- a/lib/sorcery/model/temporary_token.rb
+++ b/lib/sorcery/model/temporary_token.rb
@@ -7,12 +7,14 @@ module Sorcery
     # such as reseting password and activating the user by email.
     module TemporaryToken
       def self.included(base)
+        # FIXME: This may not be the ideal way of passing sorcery_config to generate_random_token.
+        @sorcery_config = base.sorcery_config
         base.extend(ClassMethods)
       end
 
       # Random code, used for salt and temporary tokens.
       def self.generate_random_token
-        SecureRandom.urlsafe_base64(Sorcery::Controller::Config.user_class.to_s.constantize.sorcery_config.token_randomness).tr('lIO0', 'sxyz')
+        SecureRandom.urlsafe_base64(@sorcery_config.token_randomness).tr('lIO0', 'sxyz')
       end
 
       module ClassMethods
--- a/lib/sorcery/model/temporary_token.rb
+++ b/lib/sorcery/model/temporary_token.rb
@@ -12,7 +12,7 @@ module Sorcery
 
       # Random code, used for salt and temporary tokens.
       def self.generate_random_token
-        SecureRandom.urlsafe_base64(15).tr('lIO0', 'sxyz')
+        SecureRandom.urlsafe_base64(Sorcery::Controller::Config.user_class.to_s.constantize.sorcery_config.token_randomness).tr('lIO0', 'sxyz')
       end
 
       module ClassMethods
--- a/spec/sorcery_temporary_token_spec.rb
+++ b/spec/sorcery_temporary_token_spec.rb
@@ -1,0 +1,27 @@
+require 'spec_helper'
+
+describe Sorcery::Model::TemporaryToken do
+  describe '.generate_random_token' do
+    before do
+      sorcery_reload!
+    end
+
+    subject { Sorcery::Model::TemporaryToken.generate_random_token.length }
+
+    context 'token_randomness is 3' do
+      before do
+        sorcery_model_property_set(:token_randomness, 3)
+      end
+
+      it { is_expected.to eq 4 }
+    end
+
+    context 'token_randomness is 15' do
+      before do
+        sorcery_model_property_set(:token_randomness, 15)
+      end
+
+      it { is_expected.to eq 20 }
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 ## HEAD
 
+* Add configuration for token randomness [#67](https://github.com/Sorcery/sorcery/pull/67)
+
 ## 0.11.0
 
 * Refer to User before calling remove_const to avoid NameError [#58](https://github.com/Sorcery/sorcery/pull/58)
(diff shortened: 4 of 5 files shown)


Ch4s3:
Should we set a minimum?

mtsmfm:
@Ch4s3 good to have but I'm not sure how long minimum should be 🤔

lib/sorcery/model/temporary_token.rb
```diff
@@ -12,7 +12,7 @@ def self.included(base)
 
       # Random code, used for salt and temporary tokens.
       def self.generate_random_token
-        SecureRandom.urlsafe_base64(15).tr('lIO0', 'sxyz')
+        SecureRandom.urlsafe_base64(Controller::Config.token_randomness).tr('lIO0', 'sxyz')
```
akinrt: i think this is not safety enough.
can you consider to do something like below?

```ruby
SecureRandom.urlsafe_base64([15, Controller::Config.token_randomness].max).tr('lIO0', 'sxyz')
```

akinrt: and there may be one more thing, you can calculate real length of the result value if you want.

```ruby
real_length = (15 * 3) / 4
SecureRandom.urlsafe_base64(real_length).tr('lIO0', 'sxyz').size
=> 15
```

lib/sorcery/controller/config.rb
```diff
@@ -12,6 +12,8 @@ class << self
         attr_accessor :save_return_to_url
         # set domain option for cookies
         attr_accessor :cookie_domain
+        # Set token randomness
+        attr_accessor :token_randomness
```
kyuden: `Sorcery:Controler::Config` is configuration used for `Sorcery::Controller`. But `token_randomness` is used for `Sorcery::Model::TemporaryToken` So, It is better to set it with `Sorcery:Model::Config` like [encryption_algorithm](  https://github.com/Sorcery/sorcery/blob/4e3fcdb8df16912424281322ce9f2862937846fc/lib/sorcery/model/config.rb#L44)

lib/sorcery/model/temporary_token.rb
kyuden: > [15, Controller::Config.token_randomness].max

It is weird that it isn't set the specified value by user.  So, I think the specified value should always be set.

lib/sorcery/controller/config.rb
mtsmfm: We can't access `Sorcery::Model::Config` instance directly because `Sorcery::Model::TemporaryToken` is a helper module and isn't included into `user_class`.

If you really want to do that, the code will be the following:

`Sorcery::Controller::Config.user_class.to_s.constantize.sorcery_config.token_randomness`

https://github.com/Sorcery/sorcery/blob/36698d7/lib/sorcery/adapters/active_record_adapter.rb#L22

Should I keep the setting in `Sorcery:Controler::Config` or do such trick?

lib/sorcery/model/temporary_token.rb
mtsmfm: > and there may be one more thing, you can calculate real length of the result value if you want.

I think it is difficult to control real length:

```ruby
[3] pry(main)> (15..100).select {|i| SecureRandom.urlsafe_base64((i * 3) / 4).size != i }
=> [17, 21, 25, 29, 33, 37, 41, 45, 49, 53, 57, 61, 65, 69, 73, 77, 81, 85, 89, 93, 97]
```

mtsmfm: > i think this is not safety enough.
> can you consider to do something like below?

I agree with setting minimum randomness as 15 👍

lib/sorcery/controller/config.rb
mtsmfm: 💭  IMO, it seems that `Sorcery::Model::Config` is singleton so it is not bad to access the instance as same as `Sorcery::Controller::Config`

lib/sorcery/model/temporary_token.rb
mtsmfm: We (@kyuden and I) talked in person and we think the specified value always should be respected so I won't implement minimum value.

lib/sorcery/controller/config.rb
mtsmfm: `TemporaryToken` is under `Model` module so I'll move this setting to `Sorcery::Model::Config`

mtsmfm:
@kyuden @akinrt Could you review this PR again? 🙏

lib/generators/sorcery/templates/initializer.rb
```diff
@@ -250,6 +250,12 @@
     #
     # user.subclasses_inherit_config =
 
+    # Set token randomness.
+    # The length of the result string is about 4/3 of `token_randomness`.
+    # Default: `15`
+    #
+    # config.token_randomness =
```
kyuden: s/`config.token_randomness =`/`user.token_randomness =`/

@mtsmfm  Could you actually check that not only the test code but also this setting works?

mtsmfm: @kyuden 

Fixed 🙇 

-------------

I confirmed that the token in password reset email will become longer:

screen shot 2017-06-18 at 20 37 45

screen shot 2017-06-18 at 20 38 28

kyuden:
@athix I approved this PR. Could you take a look at this?

kyuden:
@athix Are you busy in this week?

athix:
@kyuden Sorry, yeah I've been busy with some personal stuff and work. I'll try to get a review for this done as soon as I can.

kyuden:
@athix thx!

lib/sorcery/model/temporary_token.rb
```diff
@@ -12,7 +12,7 @@ def self.included(base)
 
       # Random code, used for salt and temporary tokens.
       def self.generate_random_token
-        SecureRandom.urlsafe_base64(15).tr('lIO0', 'sxyz')
+        SecureRandom.urlsafe_base64(Sorcery::Controller::Config.user_class.to_s.constantize.sorcery_config.token_randomness).tr('lIO0', 'sxyz')
```
athix: It looks like this randomization method is used globally throughout Sorcery, so I'm a bit hesitant tying the config specifically to the user. Also, it seems a bit odd that we'd have to hook through the Controller config to get the value, instead of being able to call sorcery_config directly. Looks like getting that to work isn't exactly trivial though because the generate method is outside of ClassMethods/InstanceMethods.

I'll keep digging around and see if I can figure out what's going on in there.

spec/sorcery_temporary_token_spec.rb
```diff
@@ -0,0 +1,27 @@
+require 'spec_helper'
```
athix: Yay! Thanks for writing tests for the new functionality. 👍

lib/sorcery/model/temporary_token.rb
```diff
@@ -7,12 +7,14 @@ module Model
     # such as reseting password and activating the user by email.
     module TemporaryToken
       def self.included(base)
+        # FIXME: This may not be the ideal way of passing sorcery_config to generate_random_token.
+        @sorcery_config = base.sorcery_config
```
athix: Mildly worried that this might cause `@sorcery_config` to be unintentionally exposed, but not familiar enough with how ruby inheritance/privacy works to be sure. This is how it gets saved in `lib/sorcery/model.rb` however.

athix:
@Ch4s3 I think this is probably ready to go now, mind giving this a once over?

lib/sorcery/model/temporary_token.rb
mtsmfm: `@sorcery_config` is owned by `TemporaryToken`.
It will be overridden if we include `TemporaryToken`.

```ruby
module Foo
  def self.included(base)
    @var = base.name
  end

  def self.foo
    puts @var
  end
end

class Bar
  include Foo
end

Foo.foo #=> Bar
puts Foo.instance_variable_get(:@var) #=> Bar

class Baz
  include Foo
end

Foo.foo #=> Baz
puts Foo.instance_variable_get(:@var) #=> Baz
```

Is this intentional behaviour?

athix: As far as I can tell, Sorcery::Model is the only place where TemporaryToken is/should be included. Based on your description, it sounds like it should be working exactly as intended. What I was worried about was more along the lines of `@sorcery_config` accidentally becoming available inside of a controller/model class.

For example:
```ruby
class User < ApplicationRecord
  authenticates_with_sorcery!

  def some_method
    # @sorcery_config should not be available here!
    @sorcery_config = nil # This hopefully shouldn't break things either!
  end
end
```

athix:
I'm planning on cutting a new release sometime in the next couple weeks, if I don't hear any objections before then I'll merge this in and include it for that release.

lib/sorcery/model/temporary_token.rb
mtsmfm: Sorry for late reply 💦 

Don't worry, User instance can't override `@sorcery_config`

```ruby
module Foo
  def self.included(base)
    @var = base.name
  end

  def self.foo
    @var
  end
end

class Bar
  include Foo

  def bar
    @var
  end

  def bar=(var)
    @var = var
  end
end

p Foo.foo #=> "Bar"
p Foo.instance_variable_get(:@var) #=> "Bar"
bar = Bar.new
p bar.bar #=> nil
bar.bar = :bar
p bar.bar #=> :bar
p Foo.foo #=> Bar
p Foo.instance_variable_get(:@var) #=> Bar
```

PR #68, opened 2017-06-13
Update README.md

athix:
I noticed that the contact section was out of date, so I figured I'd do some minor updates.


athix:
This shouldn't affect the gem itself, and we can always just update the README again so I'll go ahead and merge this in.

PR #76, opened 2017-07-31
Delete unsupported ruby versions from CI.

ebihara99999:
Travis CI tests ruby 2.0 and 2.1, which forces to use hash rocket. I think sorcery needs the code guideline by introducing Rubocop in the future, but this old way notation prevents it.

It's also natural to make users doubt whether this product is still maintained by supporting too old ruby versions. Of course I know sorcery is actively maintained, and it's not good.

I confirm tests are all green.

Thanks for reviewing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: ruby
 rvm:
   - jruby
-  - 2.0.0
-  - 2.1.10
   - 2.2.6
   - 2.3.3
   - 2.4.0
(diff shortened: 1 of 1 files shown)


Ch4s3:
I'm open if you are @athix

athix:
Not having to support deprecated versions sounds good to me

athix:
I also agree that we should get Rubocop going sooner rather than later. That said, still not sure what direction we want to take Sorcery from here (split into multiple gems or keep it centralized but configurable). As it is, getting the existing code fully tested and well documented should be our primary goal, IMO. Will you have any free time in the near future @Ch4s3?

ebihara99999:
>getting the existing code fully tested and well documented should be our primary goal
I agree with it. If you have tasks, I will deal with some. I'm willing to contribute to this product.

Ch4s3:
@athix I'm leaving on vacation next Thursday for 2 weeks. After that I'll have some time.

PR #80, opened 2017-08-29
Support the LINE login auth

EastResident:
I propose a new provider for LINE login authentication with auth2.0.

https://developers.line.me/line-login/overview

This PR offers only the minimal features.


Ch4s3:
Can you describe what line is?

EastResident:
@Ch4s3 
LINE is the most popular messaging application in Japan. But, it is not popular outside Japan.
LINE is similar to WhatsApp or Facebook Messenger.
About 70% of Japanese are using it, and it is used for 500 million people all over the world.

https://en.wikipedia.org/wiki/Line_(software)

Ch4s3:
@athix Should we try to push this to an external gem or roll it in?

athix:
Sorry for the late reply, I think we should probably roll in any OAuth providers that are fully ready until the gem split is complete. That said, splitting the external providers into its own gem should probably happen sooner rather than later.

@Ch4s3 perhaps we should set a time for discussion on the state of Sorcery and our near-term plans for it?

ebihara99999:
@athix
I don’t know if it’s good to propose my opinion here, three features are needed soon. At first, devise-like multi scopes’ login, which is not by STI. Every time Some I talk about Sorcery, we feel necessity of the login feature to use with more flexibility. STI is only useful for limited usages. 
Secondly, I recommend API mode support as the JWT PR implements. Sorcery will become more attractive to rails developers if Sorcery supports the api mode now that developers will make a SPA application.
The last is, to prepare full helpers needed to develop a rails app like request and  system spec helpers. Now controller specs are deprecated and recommend using system specs instead of feature specs.

Ch4s3:
@ebihara99999 and @athix let's keep this discussion moving and make a decision soon.

lion-man44:
Are there plans to merge this PR?

athix:
@lion-man44 I'll update this and merge it in when I get the chance. I think I still need to do a release for the callback issue, so this can just be rolled into that update.

lion-man44:
@athix Thank's for repling.
I looking forward to merge.

By the way, are there any workarounds?

athix:
@lion-man44 you should now be able to read off master as a work-around. Actual release coming soon™.

lion-man44:
@athix Thank you very good news!
I looking forward to it!

PR #81, opened 2017-09-06
Set user.stretches to 1 in test env by default

palkan:
Most users use BCrypt and don't change defaults, which means that the default value would be 10. That makes tests run slowly.

Setting the number of stretches in test env to 1 can decrease the test suite up to 2x times (although BCrypt minimal cost is 4, we can use 1 for SHA algorithms).


Ch4s3:
I agree, this is a nice change.

PR #82, opened 2017-09-11
add new ArgumentError for not defined user_class in config

X1ting:
I add these error, because earlier, if we forgot to define user_class (f.e. not add sorcery.rb to initializers) we got just NameError, which not easy to google for newbie developers.


Ch4s3:
Thanks @X1ting!

PR #86, opened 2017-10-07
Add QQ provider

bealking:
Add provider for QQ, which is the most used online communication tool in China.


athix:
This will be implemented via Omniauth strategies in v1.

PR #94, opened 2017-11-28
Allow user to be loaded from other source when session expires. fix #89

### LINKED ISSUE #89: User can not be loaded from other_sources when session expires in current request
I'm using such submodules:

```
[:remember_me, :reset_password, :user_activation, :brute_force_protection, :session_timeout, :activity_logging]
```

I logged in with remember_me option. And leaved tab opened for more that 1 hour. Then I reloaded page and was redirected to login page. Then I reloaded page again and I was logged in. So I started digging code and found this:

https://github.com/Sorcery/sorcery/blob/4e3fcdb8df16912424281322ce9f2862937846fc/lib/sorcery/controller/submodules/session_timeout.rb#L38
```
        base.prepend_before_action :validate_session

          def validate_session
            session_to_use = Config.session_timeout_from_last_action ? session[:last_action_time] : session[:login_time]
            if session_to_use && sorcery_session_expired?(session_to_use.to_time)
              reset_sorcery_session
              # USER IS DEFINED THERE
              @current_user = nil     
            else
              session[:last_action_time] = Time.now.in_time_zone
            end
          end
```

https://github.com/Sorcery/sorcery/blob/a873231a8c2cb997335854f18e22b066013b81b4/lib/sorcery/controller.rb#L82
```
      def current_user
        unless defined?(@current_user)
          # USER IS DEFINED SO THIS WILL BE SKIPPED IN CURRENT REQUEST
          @current_user = login_from_session || login_from_other_sources || nil
        end
        @current_user
      end
```

So maybe `remove_instance_variable :@current_user if defined? @current_user` will be better option. And IMHO this should be done everywhere where user is settled to nil. 

And also I found this place. I don't understand the reason of that ELSE block:
https://github.com/Sorcery/sorcery/blob/4e3fcdb8df16912424281322ce9f2862937846fc/lib/sorcery/controller/submodules/remember_me.rb#L70

infernalmaster:


Ch4s3:
This closes another issue right?

Ch4s3:
Thanks @infernalmaster!

Issue #87, opened 2017-10-23
`#update_attributes` uses `#update_all` inside it


ebihara99999:
`Sorcery::Adapters::ActiveRecordAdapter#update_attributes` uses `#update_all` inside instead of simply `#update`.
I think it intends one active record object to be updated because [the code](https://github.com/Sorcery/sorcery/blob/master/lib/sorcery/adapters/active_record_adapter.rb#L9) does it, so `#updated_all` is not necessary.

If I open a PR, I plan to change from 
```ruby
@model.class.where(:"#{primary_key}" => @model.send(:"#{primary_key}")).update_all(attrs)
```
to
```ruby
@model.class.find_by(:"#{primary_key}" => @model.send(:"#{primary_key}")).update(attrs)
```

The good points are:
- keep the code simple
- return boolean, not integer ( seems more understandable to me )

Thank you for watching, I would appreciate it if someone gives me suggestion.

Ch4s3:
@ebihara99999 sounds good to me.

ebihara99999:
@Ch4s3 
Thank you. I will open a PR soon!

palkan:
@Ch4s3 @ebihara99999 there are bad points too:
- now it makes 2 queries instead of one
- it invokes validations/callbacks – don't think we need them here
- it can cause race conditions.
Suppose that we have a record with `a = 2` and make two requests with different values:

thread 1 | thread 2
----------------|--------
`find_by  # loads the record with a=2`             | _
_                          |  `find_by # loads the record with a=2`
`update(a: 1)  # updates the record`     | _
_                          | `update(a: 2) # do nothing, 'cause there is no changes AR-side`

So, the final value is 1 and not 2.

Ch4s3:
@palkan interesting. We definitely want to avoid a race condition, I hadn't considered that.

palkan:
Btw, `update_columns` should do the trick: it's safe, it doesn't involve callbacks/validations and it returns `true` / `false`.

palkan:
**UPD:**  And also, what if the record has been deleted? I.e. when `find_by` returns `nil`?

Maybe, it's better to just update existing code like this:

```ruby
updated_count = @model.class.where(:"#{primary_key}" => @model.send(:"#{primary_key}")).update_all(attrs)
updated_count == 1
```

Ch4s3:
@palkan that solution to return a bool might be the way to go.

[...]

PR #98, opened 2017-12-13
Make `#update_attributes` behavior identical to `#update`

ebihara99999:
This is the PR for the [issue#87](https://github.com/Sorcery/sorcery/issues/87)
See there for details.

Another choice of implemention is like this:

```
@model.class.find_by(:"#{primary_key}" =>
@model.send(:"#{primary_key}")).update(attrs)
```

but for some reasons I don't adopt it.
First it makes 2 queries instead of one.
Secondly it has potential risk to race conditions.


Ch4s3:
closes #87

Ch4s3:
Thanks @ebihara99999!!!

PR #99, opened 2018-01-02
fix magic login migration.

masoo:
I use rails 5.1 and sorcery(github head).

so, I Migrated with this command: 
```
$ bundle exec rails g sorcery:install core magic_login --model Authentication
$ bundle exec rails db:migration
```

At that time, a migration error occurred using "magic login" module.
The problem was this part, and it succeeded when corrected.


athix:
Makes sense, and all the other migrations follow this model as well; merging.

PR #100, opened 2018-01-04
Update YARD dependency

athix:
Per [CVE-2017-17042](https://nvd.nist.gov/vuln/detail/CVE-2017-17042)


Ch4s3:
thanks @athix! I keep meaning to do this.

PR #102, opened 2018-01-18
Respect SessionTimeout on login via RememberMe

tiii:
Before this, an unlimited session was created on login via the remember me cookie, even though the SessionTimeout module is active.

I introduced a new callback (after_remember_me) to allow updating the session timeout on login from cookie without tight coupling both submodules.

It seems like all callbacks / interactions between the submodules are not tested, so I'm not sure if I should test here.. So I just added expectations for calling the specific (callback) methods.

### DIFF
--- a/lib/sorcery/controller.rb
+++ b/lib/sorcery/controller.rb
@@ -153,6 +153,10 @@ module Sorcery
         Config.after_logout.each { |c| send(c, user) }
       end
 
+      def after_remember_me!(user)
+        Config.after_remember_me.each { |c| send(c, user) }
+      end
+
       def user_class
         @user_class ||= Config.user_class.to_s.constantize
       rescue NameError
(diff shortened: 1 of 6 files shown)


athix:
Adding as much testing as you can would be appreciated, Sorcery's test suite is woefully under where I would like it to be.

Ch4s3:
Yeah, I agree with @athix we definitely need more tests for this. I like the direction otherwise, and will be happy to merge it once we're sure it's well covered.

tiii:
I don't have time to write the missing tests for the whole gem. By the time I submitted this, it was the only callback that had tests at all. So you are requesting that I come up with a concept on how to test callbacks in general for sorcery?

If there are any other callback tests in the meantime i'd be happy to adapt them to the current use case.

Ch4s3:
@tiii You have a good point. I'm going to merge this and we can use this as a jumping off point.

PR #107, opened 2018-02-03
Add rubocop config and todo

athix:
I have the feeling this will differ significantly from our current codacy settings, so might need to tweak the configs a little before merging. Once merged, we should be able to tell codacy to read from the main config.

Required to fix error on #57


athix:
Looks like codacy can't read the config unless it's merged into master first. I guess I'll just plan on opening a new PR if it causes issues. ¯\\\_(ツ)\_/¯

PR #110, opened 2018-02-21
Add an opt in invalidate_active_sessions!  method to session timeout

fangbyte:
Add a configurable `invalid_active_sessions!` method to the session_timeouts module. Configurable, but expects an `invalidate_sessions_before` timestamp column on user. Does not invalidate the session if `invalidate_sessions_before` is not set meaning that if it is deployed it won't log out currently logged in users.

Update the gem version if necessary. Because this is opt in, it should not be a breaking change.

@arnvald

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,12 @@ User.load_from_reset_password_token(token)
 @user.change_password!(new_password)
 ```
 
+### Session Timeout
+
+```ruby
+invalidate_active_sessions! #Invalidate all sessions with a login_time or last_action_time before the current time. Must Opt-in
+```
+
 ### User Activation
 
 ```ruby
@@ -184,6 +190,7 @@ Inside the initializer, the comments will tell you what each setting does.
 
 - Configurable session timeout
 - Optionally session timeout will be calculated from last user action
+- Optionally enable a method to clear all active sessions, expects an `invalidate_sessions_before` datetime attribute.
 
 **Brute Force Protection** (see [lib/sorcery/model/submodules/brute_force_protection.rb](https://github.com/Sorcery/sorcery/blob/master/lib/sorcery/model/submodules/brute_force_protection.rb)):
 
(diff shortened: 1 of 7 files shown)


Ch4s3:
This looks pretty good. Can you update the readme?

arnvald:
@fangbyte @Ch4s3 I am not 100% sure if I'm right, but I'm afraid this concept cannot work. Let me explain it in 2 parts:

1. Current situation - the `invalidate_sessions_before` is never cleared, which means that once I try to invalidate all sessions, I will never be able to log in again. Or rather, I will be able to log in just to be logged out on the next requests.

The only way to fix I can think of is to clear the `invalidate_sessions_before` after I log in. However that causes another big issue:

2. With fix - since `invalidate_sessions_before` is cleared after log in, a following scenario can happen

* I leave my account logged in on a public computer
* at home, I remember that I left it there, so I click "log out all sessions", this makes the `invalidate_sessions_before`
* still at home, I want to use the application again, so I log in, therefore `invalidate_sessions_before` value is cleared
* on that public computer, someone opens the website. Because the `invalidate_sessions_before` was cleared, the session is not invalidated, so someone can use my account.

fangbyte:
@arnvald I don't think you are logged out after the current session, since the `invalidate_sessions_before` timestamp is checked against the session `login_time` if it is present or `last_action_time` if login time is not present (for sessions created with the remember me token). Any sessions logged in _after_ the `invalidate_sessions_before` timestamp should not be invalidated because `login_time` or `last_action_time` will be greater than the `invalidate_sessions_before` timestamp. I have added a test for the situation you mentioned in 1 in e9f802b

arnvald:
@fangbyte you're absolutely right, I misunderstood that part. The code will work as intended. Thank you for explanation!

fangbyte:
@Ch4s3 @arnvald  I have updated the readme

Let me know if there's anything else I can do to move this forward.

fangbyte:
I pushed two additional commits fixing small issues: 
* Move `invalidate_active_sessions!` to no longer be a protected method. It should never have been protected.
* Added a default of `Time.now.in_time_zone` to the check to invalidate a session if `login_time` and `last_action_time` are not set rather than allowing it to raise a no method error on nil if those session values are not set.

Ch4s3:
Awesome! @fangbyte thanks so much for your patience!

PR #112, opened 2018-03-12
update oauth2 gem dependency

FunkyloverOne:


sorcery.gemspec
```diff
@@ -21,7 +21,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.2.2'
 
   s.add_dependency 'oauth', '~> 0.4', '>= 0.4.4'
-  s.add_dependency 'oauth2', '~> 1.0', '>= 0.8.0'
```
athix: `~> 1.0` should already be locking to anything greater than 1.0 and less than 2.0, so using `~> 1.4` should be unnecessary. You can also lock your project's Gemfile to a particular version of oauth2, and as long as it's within this range there shouldn't be any conflicts. (e.g. `gem 'oauth2', '~> 1.4.0'` inside `./Gemfile`)

If this is due to a security vulnerability or similar, we should probably update the hard limit instead (e.g. `'>= 1.4.0'`)

FunkyloverOne: @athix you're right, it doesn't seem to be any security vulnerabilities, I just was struggling with version conflicts in my project. But I made this PR because I thought that it might be a good idea to use fresher version, but most likely not. Closing this one.
P.S. thanks for a tip with explicit Gemfile dependencies :)

PR #130, opened 2018-06-13
send activation email after the commit in transactional databases

anaumov:
Use `after_commit on: :create` instead of `after_create` callback for sending activation email. It's necessary because of `after_create` is located in a transaction. There is no guarantee that user will exist when email will start building.

More details http://codebeerstartups.com/2012/11/use-after_commit-instead-of-active-record-callbacks-to-avoid-unexpected-errors/

Also added a patch to mongo `define callback`. Let me know if you need more info about this issue.

### DIFF
--- a/lib/sorcery/adapters/mongoid_adapter.rb
+++ b/lib/sorcery/adapters/mongoid_adapter.rb
@@ -33,7 +33,15 @@ module Sorcery
         end
 
         def define_callback(time, event, method_name, options={})
-          @klass.send "#{time}_#{event}", method_name, options.slice(:if)
+          callback_name = if event == :commit && options[:on] == :create
+                            "#{time}_create"
+                          elsif event == :commit
+                            "#{time}_save"
+                          else
+                            "#{time}_#{event}"
+                          end
+
+          @klass.send callback_name, method_name, options.slice(:if)
         end
 
         def credential_regex(credential)
--- a/lib/sorcery/adapters/mongoid_adapter.rb
+++ b/lib/sorcery/adapters/mongoid_adapter.rb
@@ -33,10 +33,10 @@ module Sorcery
         end
 
         def define_callback(time, event, method_name, options={})
-          @klass.send callback_name(type, event, options), method_name, options.slice(:if)
+          @klass.send callback_name(time, event, options), method_name, options.slice(:if)
         end
 
-        def callback_name(type, event, options)
+        def callback_name(time, event, options)
           if event == :commit
             options[:on] == :create ? "#{time}_create" : "#{time}_save"
           else
--- a/lib/sorcery/adapters/mongoid_adapter.rb
+++ b/lib/sorcery/adapters/mongoid_adapter.rb
@@ -33,15 +33,15 @@ module Sorcery
         end
 
         def define_callback(time, event, method_name, options={})
-          callback_name = if event == :commit && options[:on] == :create
-                            "#{time}_create"
-                          elsif event == :commit
-                            "#{time}_save"
-                          else
-                            "#{time}_#{event}"
-                          end
+          @klass.send callback_name(type, event, options), method_name, options.slice(:if)
+        end
 
-          @klass.send callback_name, method_name, options.slice(:if)
+        def callback_name(type, event, options)
+          if event == :commit
+            options[:on] == :create ? "#{time}_create" : "#{time}_save"
+          else
+            "#{time}_#{event}"
+          end
         end
 
         def credential_regex(credential)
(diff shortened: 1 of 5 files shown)


lib/sorcery/adapters/mongoid_adapter.rb
```diff
@@ -33,7 +33,15 @@ def define_field(name, type, options={})
         end
 
         def define_callback(time, event, method_name, options={})
-          @klass.send "#{time}_#{event}", method_name, options.slice(:if)
+          callback_name = if event == :commit && options[:on] == :create
```
youzik: it's better to exclude it to a method and use case instead of if/elsif

anaumov:
@youzik thanks for the comment, I've moved code to separate method. Not sure about the case statement, because a condition does not depend on one parameter.

Can you suggest me how to reproduce specs failing locally? https://travis-ci.org/Sorcery/sorcery/jobs/391800300 . 

I tried:
```
export BUNDLE_GEMFILE=$PWD/gemfiles/active_record-rails41.gemfile
bundle install
bundle exec rake
/Users/naumov/.rbenv/versions/2.4.0/bin/ruby -I/Users/naumov/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/rspec-core-3.7.1/lib:/Users/naumov/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/rspec-support-3.7.1/lib /Users/naumov/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/rspec-core-3.7.1/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb
/Users/naumov/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activesupport-4.1.16/lib/active_support/core_ext/numeric/conversions.rb:121: warning: constant ::Fixnum is deprecated
/Users/naumov/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/activesupport-4.1.16/lib/active_support/core_ext/numeric/conversions.rb:121: warning: constant ::Bignum is deprecated
/Users/naumov/.rbenv/versions/2.4.0/bin/ruby -I/Users/naumov/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/rspec-core-3.7.1/lib:/Users/naumov/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/rspec-support-3.7.1/lib /Users/naumov/.rbenv/versions/2.4.0/lib/ruby/gems/2.4.0/gems/rspec-core-3.7.1/exe/rspec --pattern spec/\*\*\{,/\*/\*\*\}/\*_spec.rb failed
```

athix:
@anaumov I believe the issue you're having is that Fixnum and Bignum were refactored in ruby itself in either `2.3` or `2.4`

Try switching rvm to use the same version as the failing travis build: `ruby 2.2.9p480`

anaumov:
@athix I don't care about Fixnum and Bignum issues :) I worry that next line fails without any backtrace
```

**Comment by anaumov** [comment]
I turned off `use_transactional_fixtures` in specs in order to be able to test `after_comit` callback in rails 4. Specs execution time are still the same. Let me know if you need any cleanup in PR.

 Thanks @athix, I set ruby to 2.9 and specs work fine.

File: lib/sorcery/model/submodules/user_activation.rb [hunk 2]
```diff
@@ -45,7 +45,7 @@ def self.included(base)
             # don't setup activation if no password supplied - this user is created automatically
             sorcery_adapter.define_callback :before, :create, :setup_activation, if: proc { |user| user.send(sorcery_config.password_attribute_name).present? }
             # don't send activation needed email if no crypted password created - this user is external (OAuth etc.)
-            sorcery_adapter.define_callback :after, :create, :send_activation_needed_email!, if: :send_activation_needed_email?
+            sorcery_adapter.define_callback :after, :commit, :send_activation_needed_email!, on: :create, if: :send_activation_needed_email?
```
athix: Interesting, I honestly didn't realize rails had a callback for specifically after the DB commit.

:+1:

spec/spec_helper.rb
```diff
@@ -29,7 +29,7 @@ class TestMailer < ActionMailer::Base; end
   config.include RSpec::Rails::ControllerExampleGroup, file_path: /controller(.)*_spec.rb$/
   config.mock_with :rspec
 
-  config.use_transactional_fixtures = true
+  config.use_transactional_fixtures = false
```
athix: @anaumov is this to fix the callbacks not working due to it waiting for a commit that never happens? I'd be worried that this will break specs for end-users as well.

anaumov: Yes, if it sets to `use_transactional_fixtures  = true` specs running with `rails  ~> 4.0` didn't commit transaction. No emails will be sent and two specs are failed. `use_transactional_fixtures  = false` works fine (the build is green). I read that it could slow down specs, but they are still pretty fast.

Ch4s3:
Ok, looks good to me. Might fix #134?

anaumov:
Cool! thanks @Ch4s3

PR #133, opened 2018-06-21
Support injecting helpers into ActionController::API

andrewhood125:
Rails API apps use the lighter alternative to ActionController::Base.
This adds support for injecting these helpers into either action
controller base class that the consuming application happens to inherit
from.

This is a step towards full support in rails api mode.

### DIFF
--- a/lib/sorcery/engine.rb
+++ b/lib/sorcery/engine.rb
@@ -8,9 +8,11 @@ module Sorcery
     config.sorcery = ::Sorcery::Controller::Config
 
     initializer 'extend Controller with sorcery' do
-      ActionController::Base.send(:include, Sorcery::Controller)
-      ActionController::Base.helper_method :current_user
-      ActionController::Base.helper_method :logged_in?
+      ActiveSupport.on_load('action_controller') do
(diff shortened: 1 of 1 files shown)


lib/sorcery/engine.rb
```diff
@@ -8,9 +8,11 @@ class Engine < Rails::Engine
     config.sorcery = ::Sorcery::Controller::Config
 
     initializer 'extend Controller with sorcery' do
-      ActionController::Base.send(:include, Sorcery::Controller)
-      ActionController::Base.helper_method :current_user
-      ActionController::Base.helper_method :logged_in?
+      ActiveSupport.on_load('action_controller') do
+        self.send(:include, Sorcery::Controller)
```
youzik: `self.` is redundant

andrewhood125: Thanks!

athix:
It appears this broke master:

```ruby
NoMethodError: undefined method `external_providers' for Sorcery::Controller::Config:Module
```

Opening a PR to fix this.

PR #135, opened 2018-07-09
Revert "Namespace class method #encrypt"

athix:
#77 was merged with breaking CI, and should be fixed or reverted.

### DIFF
--- a/lib/sorcery/model.rb
+++ b/lib/sorcery/model.rb
@@ -119,7 +119,7 @@ module Sorcery
       end
 
       # encrypt tokens using current encryption_provider.
-      def sorcery_encrypt(*tokens)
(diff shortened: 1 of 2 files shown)


athix:
Looks like the failure was from a spec that got skipped somehow, but I can't retry the build as its already been merged and fails permanently now (fails to fetch). Will close if the rebuild on master passes.

athix:
The jruby stuff is ignored for build pass/fail, looks like the build on master did pass though so it was probably just a glitch in Travis. I would definitely try to rebuild before merging if we think it's just a glitch though; better safe than sorry.

Ch4s3:
I'll take a look later.

PR #139, opened 2018-08-09
Remove `publish_actions` permission for facebook

hetakan:
In facebook, `publish_actions` permission has been deleted.
https://developers.facebook.com/docs/graph-api/changelog/breaking-changes?locale=en_US

So, it is not good that `publish_actions` still exists in template file.


lib/generators/sorcery/templates/initializer.rb
```diff
@@ -119,7 +119,7 @@
   # config.facebook.callback_url = "http://0.0.0.0:3000/oauth/callback?provider=facebook"
   # config.facebook.user_info_path = "me?fields=email"
   # config.facebook.user_info_mapping = {:email => "email"}
-  # config.facebook.access_permissions = ["email", "publish_actions"]
+  # config.facebook.access_permissions = ["email"]
```
athix: Confirmed permission was removed with no replacement.

PR #142, opened 2018-08-18
fix CipherError class name

ryochin:


lib/sorcery/crypto_providers/aes256.rb
```diff
@@ -29,7 +29,7 @@ def encrypt(*tokens)
 
         def matches?(crypted, *tokens)
           decrypt(crypted) == tokens.join
-        rescue OpenSSL::CipherError
+        rescue OpenSSL::Cipher::CipherError
```
athix: Confirmed that this appears to be a typo. I am curious why it hasn't been discovered until now however. Shouldn't this be propagating up the stack every time a login fails?

PR #153, opened 2018-10-19
Clean up initializer comments

valentine:
I've tried to make the documentation in the `initializer.rb` file (which gets copied to `config/sorcery.rb` on install) clearer for new users. 

I've also added the `magic_login` submodule to the list, since that seems like a recent addition.

### DIFF
--- a/lib/generators/sorcery/templates/initializer.rb
+++ b/lib/generators/sorcery/templates/initializer.rb
@@ -1,5 +1,6 @@
 # The first thing you need to configure is which modules you need in your app.
 # The default is nothing which will include only core features (password encryption, login/logout).
+#
 # Available submodules are: :user_activation, :http_basic_auth, :remember_me,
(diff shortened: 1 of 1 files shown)


athix:
Thanks for your help @valentine! I'll review this when I have some time and merge it in. Any other help with the documentation is greatly appreciated. :)

lib/generators/sorcery/templates/initializer.rb
```diff
@@ -502,23 +503,23 @@
...
+    # User's external unique identifier in the `authentications` class.
     # Default: `:uid`
     #
     # user.provider_uid_attribute_name =
   end
 
   # This line must come after the 'user config' block.
   # Define which model authenticates with sorcery.
-  config.user_class = '<%= model_class_name %>'
+  config.user_class = "<%= model_class_name %>"
```
athix: Personally I prefer single quotes for strings that do not require interpolation. Not a blocking issue however.

PR #158, opened 2018-11-08
Add openid connect provider and allow to attach query params

obfuscoder:
I was trying to integrate an openid connect provider with any of the existing external providers. None of which is exactly supporting the standard.

With this change an external provider "openid" is added which is compatible with any provider implementing the standard openid connect protocol. In order to allow using additional parameters defined in openid conenct the oauth2 module was extended.

Let me know what you think of this. If this is something you would like to pull, I can enhance with test code and documentation.

### DIFF
--- a/lib/sorcery/protocols/oauth2.rb
+++ b/lib/sorcery/protocols/oauth2.rb
@@ -8,13 +8,16 @@ module Sorcery
       end
 
       def authorize_url(options = {})
-        client = build_client(options)
-        client.auth_code.authorize_url(
+        query_params = {
           redirect_uri: @callback_url,
           scope: @scope,
           display: @display,
           state: @state
-        )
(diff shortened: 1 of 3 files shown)


lib/sorcery/protocols/oauth2.rb
```diff
@@ -8,13 +8,16 @@ def oauth_version
       end
 
       def authorize_url(options = {})
-        client = build_client(options)
-        client.auth_code.authorize_url(
+        query_params = {
```
athix: Why are we adding the query parameters to the options hash here? It doesn't seem to affect any other providers, and if it's specific to the generic openid provider then it should probably be encapsulated there.

athix:
Please reopen if there's renewed interest in this.

PR #124, opened 2018-05-16
Changed VK provider uid key

bragovo:
Since version 5.0 https://vk.com/dev/version5 API returns ID instead UID. UID working only with old 4.x versions.

PR #160, opened 2018-11-20
Add Auth0 as external provider

athix:


athix:
Interesting...Tests pass locally, and these are the same errors as #124.

athix:
Turns out the hash we stub out was missing `sub` for the UID. Should be the same issue for the VK pr.

PR #163, opened 2018-12-10
update_failed_logins_count is called twice when login failed

kissrobber:
Failed_logins_count is incremented by 2 when login failed.

login failed request SQL log
```
CACHE User Load (0.0ms)  SELECT  `users`.* FROM `users` WHERE `users`.`type` IN ('User') AND `users`.`deleted_at` IS NULL AND `users`.`email` = 'unko@example.com' ORDER BY `users`.`id` ASC LIMIT 1  [["LIMIT", 1]]
SQL (5.8ms)  UPDATE `users` SET `failed_logins_count` = COALESCE(`failed_logins_count`, 0) + 1 WHERE `users`.`type` IN ('User') AND `users`.`id` = 1702
User Load (0.7ms)  SELECT  `users`.* FROM `users` WHERE `users`.`type` IN ('User') AND `users`.`deleted_at` IS NULL AND `users`.`email` = 'unko@example.com' ORDER BY `users`.`id` ASC LIMIT 1
SQL (3.2ms)  UPDATE `users` SET `failed_logins_count` = COALESCE(`failed_logins_count`, 0) + 1 WHERE `users`.`type` IN ('User') AND `users`.`id` = 1702
```

Config.after_failed_login and Config.after_login are set doubly
```
(byebug) Rails.application.config.sorcery.submodules
[:user_activation, :reset_password, :activity_logging, :brute_force_protection, :external]
(byebug) Sorcery::Controller::Config.after_login
[:register_login_time_to_db, :register_last_ip_address, :reset_failed_logins_count!, :register_login_time_to_db, :register_last_ip_address, :reset_failed_logins_count!]
(byebug) Sorcery::Controller::Config.after_failed_login
[:update_failed_logins_count!, :update_failed_logins_count!]
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,6 @@ source 'https://rubygems.org'
 gem 'pry'
 gem 'rails', '~> 5.2.0'
 gem 'rails-controller-testing'
-gem 'sqlite3'
+gem 'sqlite3', '~> 1.3.6'
 
 gemspec
--- a/lib/sorcery/controller/submodules/activity_logging.rb
+++ b/lib/sorcery/controller/submodules/activity_logging.rb
@@ -30,9 +30,16 @@ module Sorcery
             end
             merge_activity_logging_defaults!
           end
-          Config.after_login    << :register_login_time_to_db
-          Config.after_login    << :register_last_ip_address
-          Config.before_logout  << :register_logout_time_to_db
+          # FIXME: There is likely a more elegant way to safeguard these callbacks.
+          unless Config.after_login.include?(:register_login_time_to_db)
+            Config.after_login << :register_login_time_to_db
+          end
+          unless Config.after_login.include?(:register_last_ip_address)
+            Config.after_login << :register_last_ip_address
+          end
+          unless Config.before_logout.include?(:register_logout_time_to_db)
+            Config.before_logout << :register_logout_time_to_db
+          end
           base.after_action :register_last_activity_time_to_db
         end
 
--- a/lib/sorcery/controller/submodules/brute_force_protection.rb
+++ b/lib/sorcery/controller/submodules/brute_force_protection.rb
@@ -10,9 +10,13 @@ module Sorcery
       module BruteForceProtection
         def self.included(base)
           base.send(:include, InstanceMethods)
-
-          Config.after_login << :reset_failed_logins_count!
-          Config.after_failed_login << :update_failed_logins_count!
+          # FIXME: There is likely a more elegant way to safeguard these callbacks.
+          unless Config.after_login.include?(:reset_failed_logins_count!)
+            Config.after_login << :reset_failed_logins_count!
+          end
+          unless Config.after_failed_login.include?(:update_failed_logins_count!)
+            Config.after_failed_login << :update_failed_logins_count!
+          end
         end
 
         module InstanceMethods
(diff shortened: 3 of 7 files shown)


athix:
I was under the impression that `ActionController::Base` and `ActionController::API` didn't inherit from each other in any way. Seems weird that including the controller in each separately would be the cause of a callback duplication.

@kissrobber would you mind helping me understand what's going wrong here?

kissrobber:
in rails
Both `defined?(ActionController::API)` and `defined?(ActionController::Base)` are true.
https://github.com/rails/rails/tree/master/actionpack/lib/action_controller

in sorcery
Both `ActionController::API.send(:include, Sorcery::Controller)` and `ActionController::Base.send(:include, Sorcery::Controller)` fire `Sorcery::Controller.included`
https://github.com/Sorcery/sorcery/blob/master/lib/sorcery/controller.rb#L3
↓
https://github.com/Sorcery/sorcery/blob/master/lib/sorcery/controller/submodules/brute_force_protection.rb#L11

kissrobber:
> I was under the impression that `ActionController::Base` and `ActionController::API` didn't inherit from each other in any way. Seems weird that including the controller in each separately would be the cause of a callback duplication.
> 
> @kissrobber would you mind helping me understand what's going wrong here?

kissrobber:
Sorry I closed this PR by mistake.

athix:
@kissrobber That shouldn't cause a duplication issue though, API and Base don't inherit each other as far as I can tell. This change would break applications that use API for a subset of controllers, and Base for the rest.

Can you provide a minimal example controller so I can replicate this issue and try it out?

kissrobber:
> API and Base don't inherit each other

That's not a problem here.

In the original code
Both the if conditions are evaluated as `true` regardless whether they inherit each other or not.
```
      if defined?(ActionController::API) # -> true
        ActionController::API.send(:include, Sorcery::Controller) 
      end

      if defined?(ActionController::Base) # -> true
        ActionController::Base.send(:include, Sorcery::Controller)
        ActionController::Base.helper_method :current_user
        ActionController::Base.helper_method :logged_in?
      end
```
Then `ActionController::Base.send(:include, Sorcery::Controller)` is called twice always.

So, this called twice
```
          Config.after_login << :reset_failed_logins_count!
          Config.after_failed_login << :update_failed_logins_count!
```
https://github.com/Sorcery/sorcery/blob/master/lib/sorcery/controller/submodules/brute_force_protection.rb#L11

> Can you provide a minimal example controller so I can replicate this issue and try it out?

You mean it has been confirmed that this issue doesn't occur in your code?
In that case, I will do that sometime soon 👍

athix:
Ahh, I understand the issue now. I thought it was some weird inheritance, but it's just the `self.included` code not safe-guarding against being included by different controllers. You're right, although I think the better solution here is updating the `self.included` method to not double down if it gets included multiple times. I'll also double check the others and fix those as well.

Don't worry about the example, I was just misunderstanding the problem.

athix:
Closes #172

petervanwesep:
Is @kissrobber's solution a reasonable monkey patch for this issue? Need a stopgap upgrade in response to CVE-2019-5418 and CVE-2019-5419. Thanks!

athix:
Sorry for the delay, will merge this into master after CI passes.

Not fully convinced this solves all the potential issues with double inclusion, but I confirmed locally that this prevents the double callbacks for the controller submodules.

PR #165, opened 2018-12-16
Add #change_password method to reset_password module.

SeriouslyAwesome:
Serves the same function as `#change_password!` but without raising exceptions on save. Useful for returning validation errors back to the view, for example, when `:password_confirmation` did not match `:password`.

### DIFF
--- a/lib/sorcery/model/submodules/reset_password.rb
+++ b/lib/sorcery/model/submodules/reset_password.rb
@@ -124,10 +124,14 @@ module Sorcery
           end
 
           # Clears token and tries to update the new password for the user.
-          def change_password!(new_password)
+          def change_password(new_password, raise_on_failure: false)
             clear_reset_password_token
             send(:"#{sorcery_config.password_attribute_name}=", new_password)
-            sorcery_adapter.save raise_on_failure: true
+            sorcery_adapter.save raise_on_failure: raise_on_failure
+          end
+
+          def change_password!(new_password)
+            change_password(new_password, raise_on_failure: true)
           end
 
           protected
--- a/spec/shared_examples/user_reset_password_shared_examples.rb
+++ b/spec/shared_examples/user_reset_password_shared_examples.rb
@@ -14,6 +14,8 @@ shared_examples_for 'rails_3_reset_password_model' do
     context 'API' do
       specify { expect(user).to respond_to :deliver_reset_password_instructions! }
 
+      specify { expect(user).to respond_to :change_password }
+
       specify { expect(user).to respond_to :change_password! }
 
       it 'responds to .load_from_reset_password_token' do
@@ -314,15 +316,25 @@ shared_examples_for 'rails_3_reset_password_model' do
       end
     end
 
-    it 'when change_password! is called, deletes reset_password_token' do
+    it 'when change_password is called, deletes reset_password_token' do
       user.deliver_reset_password_instructions!
 
       expect(user.reset_password_token).not_to be_nil
 
-      user.change_password!('blabulsdf')
-      user.save!
+      user.change_password('blabulsdf')
 
       expect(user.reset_password_token).to be_nil
+    end
+
+    it 'when change_password! is called, calls change_password' do
+      new_password = 'blabulsdf'
+
+      user.deliver_reset_password_instructions!
+      expect(user.reset_password_token).not_to be_nil
+
+      expect(user).to receive(:change_password).with(new_password, raise_on_failure: true)
+
+      user.change_password!(new_password)
     end
 
     it 'returns false if time between emails has not passed since last email' do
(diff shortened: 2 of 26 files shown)


SeriouslyAwesome:
The Travis build seems to be erroring out for reasons unrelated to my changes, but specs pass locally on 2.3. Let me know if there's something I'm missing to get the build passing in CI.

athix:
Huh, looks like RVM made some changes that might have broken our builds. I'll look into it and rebuild once it's fixed.

athix:
@SeriouslyAwesome update on the CI issue. It looks like RVM added a new maintainer signing key which broke all our builds. TravisCI has a fix, but it hasn't been deployed yet. Once it's deployed, this will probably pass on rebuild.

SeriouslyAwesome:
Thanks @athix! Let me know if anything else is needed for this to be merged.

lib/sorcery/model/submodules/reset_password.rb
```diff
@@ -124,10 +124,14 @@ def reset_password_reset_page_access_counter
           end
 
           # Clears token and tries to update the new password for the user.
-          def change_password!(new_password)
+          def change_password(new_password, raise_on_failure: false)
```
athix: For default values, I've normally seen `def method(variable = value)` rather than `def method(variable: value)`, are you sure this works as intended?

spec/shared_examples/user_reset_password_shared_examples.rb
```diff
@@ -314,17 +316,27 @@
       end
     end
 
-    it 'when change_password! is called, deletes reset_password_token' do
+    it 'when change_password is called, deletes reset_password_token' do
```
athix: This technically changes test functionality, I think leaving it as `change_password!` here is reasonable.

spec/shared_examples/user_reset_password_shared_examples.rb
```diff
@@ -314,17 +316,27 @@
...
       user.deliver_reset_password_instructions!
 
       expect(user.reset_password_token).not_to be_nil
 
-      user.change_password!('blabulsdf')
-      user.save!
+      user.change_password('blabulsdf')
```
athix: Same as above, bang method should be fine here. Interesting that `user.save!` was used but isn't necessary. I should probably look into if that's intentional.

spec/shared_examples/user_reset_password_shared_examples.rb
```diff
@@ -314,17 +316,27 @@
...
 
       expect(user.reset_password_token).to be_nil
     end
 
+    it 'when change_password! is called, calls change_password' do
```
athix: This test seems a bit out of place. Testing that both the normal and bang method work makes sense, but testing that one calls the other doesn't seem beneficial.

SeriouslyAwesome:
Hey @athix thanks for the review. You made some good points about the tests, so I tweaked those to more explicitly test both methods and the differences between them. Also see my comment about the `save_on_failure: false` argument. Ready for another look at your convenience.

spec/shared_examples/user_reset_password_shared_examples.rb
```diff
@@ -314,13 +316,25 @@
...
       expect(user.reset_password_token).not_to be_nil
+      expect(user).to receive(:save!)
 
       user.change_password!('blabulsdf')
-      user.save!
+
+      expect(user.reset_password_token).to be_nil
+    end
+
+    it 'when change_password! is called, deletes reset_password_token but does not save record' do
```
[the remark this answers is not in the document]
SeriouslyAwesome: Doh! Yep.

spec/shared_examples/user_reset_password_shared_examples.rb
```diff
@@ -314,13 +316,25 @@
...
+      new_password = 'blabulsdf'
+
+      user.deliver_reset_password_instructions!
+      expect(user.reset_password_token).not_to be_nil
+      expect(user).to_not receive(:save!)
```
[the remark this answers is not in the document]
SeriouslyAwesome: Wow, yeah. Not sure what I was thinking when I wrote those descriptions... I've clarified the language and also updated the tests to specifically check that either `#save` or `#save!` is being called and that the other is not being called.

SeriouslyAwesome:
Okay @athix, hopefully things should make more sense now. Apologies for the confusion... apparently got too deep inside my own head late one night while writing this.

I'm not sure what's going on with the build though. Seems to be related to the recent dropping of Rails 4.X support, but Travis is still looking for `sqlite`. If there's anything I can do on my side to fix that, let me know.

athix:
I recently pushed a fix for sqlite3 breaking, I'll try merging master into this branch and see if TravisCI plays ball.

spec/shared_examples/user_reset_password_shared_examples.rb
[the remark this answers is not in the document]
athix: Haha, I feel you. Thanks for updating! I'll give this another pass after CI passes and merge if it all looks good.

PR #166, opened 2019-01-12
Fix `rvm get stable` breaking due to new maintainer signature

athix:
See related issues:
[rvm/rvm#4520](https://github.com/rvm/rvm/issues/4520)
[rvm/rvm#4561](https://github.com/rvm/rvm/issues/4561)

Should be solved by (but isn't yet):
[travis-ci/travis-build#1634](https://github.com/travis-ci/travis-build/pull/1634)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,6 @@ before_script:
   - mysql -e 'create database sorcery_test;'
 
 before_install:
-  - gpg --keyserver hkp://pool.sks-keyservers.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
-  - rvm get stable --auto-dotfiles
   - gem update bundler
 
 matrix:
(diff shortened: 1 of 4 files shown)


athix:
Somehow this breaks installing bundler? 😕

.travis.yml
```diff
@@ -15,6 +15,7 @@ before_script:
   - mysql -e 'create database sorcery_test;'
 
 before_install:
+  - gpg --keyserver hkp://pool.sks-keyservers.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
   - rvm get stable --auto-dotfiles
```
eregon: Why are you updating rvm here? AFAIK the `rvm:` above runs before `before_install` and installs Ruby versions before this line is run.

eregon: It is the case as shown in https://travis-ci.org/Sorcery/sorcery/jobs/478627382
Seems better now, but looks like Rails depends on `bundler < 2`.

athix:
Spending too much time trying to solve this, I think it makes sense to just wait for the TravisCI fix to deploy.

PR #70, opened 2017-06-18
JWT authentication

wilddima:
Hi!
I implemented JWT authentication as module and tested it works capacity in my pet project(SPA with React). Maybe, configurations is looking complex, but i tried to give more flexibility for customization authentication. I also tried to write some documentation, but my english is not good enough for it.

PR #167, opened 2019-01-12
JWT authentication ( after the commit b06aa7a )

ebihara99999:
This is from https://github.com/Sorcery/sorcery/pull/70.
There are some reviews with no response and the PR is stale for a long time (though the PR is great!!).

Thank @wilddima for implementing them firstly!!


ebihara99999:
TODO

- the discussion: https://github.com/Sorcery/sorcery/pull/70#discussion_r247848379
- the discussion: https://github.com/Sorcery/sorcery/pull/70#discussion_r247848621
- the discussion https://github.com/Sorcery/sorcery/pull/70#discussion_r247316356

ebihara99999:
@athix Thanks, I've fixed them just now

sorcery.gemspec
```diff
@@ -37,6 +37,8 @@ Gem::Specification.new do |s|
   s.add_dependency 'bcrypt', '~> 3.1'
   s.add_dependency 'oauth', '~> 0.4', '>= 0.4.4'
   s.add_dependency 'oauth2', '~> 1.0', '>= 0.8.0'
+  s.add_dependency 'bcrypt', '~> 3.1'
```
athix: @ebihara99999 Duplicate `bcrypt` dependency, likely from a merge. I would double check anything that changed on your merge commits.

ebihara99999: @athix I'm sorry to add the wrong dependency. I've fixed it just now

athix:
If anyone in the community would like to help with this, I think JWT support would be a great addition to Sorcery.

lib/generators/sorcery/templates/initializer.rb
```diff
@@ -523,6 +523,43 @@
...
+
+    # Key on which returned token
+    # Default: :auth_token
+    #
+    # config.jwt_auth_token_key =
+
+    # A flag that specifies whether to perform a database query to set the current_user
+    # Default: true
+    #
+    # config.jwt_set_user =
```
mladenilic: Do we need this flag? Is there ever a case when `current_user` should not be an instance of a user model?

lib/sorcery/controller/submodules/jwt_auth.rb
```diff
@@ -0,0 +1,85 @@
...
+              payload = default_payload.merge Config.jwt_payload
+              
+              { Config.jwt_user_data_key => default_payload,
+                Config.jwt_auth_token_key => jwt_encode(payload) }
+            end
+          end
+
+          # To be used as a before_action.
+          def jwt_require_auth
+            binding.pry
```
mladenilic: I guess this is here by accident? :)

lib/sorcery/controller/submodules/jwt_auth.rb
```diff
@@ -0,0 +1,85 @@
...
+  module Controller
+    module Submodules
+      module JwtAuth
+        def self.included(base)
+          base.send(:include, InstanceMethods)
+        end
+
+        module InstanceMethods
+          # This method return generated token if user can be authenticated
+          def jwt_auth(*credentials)
```
mladenilic: I think slightly more appropriate name for the method would be `jwt_authenticate`. Simply to stay consistent with `User.authenticate`.

mladenilic: Method should also allow users to get login failure reason, similar to what `login` method does.

lib/sorcery/controller/submodules/jwt_auth.rb
```diff
@@ -0,0 +1,85 @@
...
+                Config.jwt_auth_token_key => jwt_encode(payload) }
+            end
+          end
+
+          # To be used as a before_action.
+          def jwt_require_auth
+            binding.pry
+            @current_user = Config.jwt_set_user ? User.find(jwt_user_id) : jwt_user_data
+          rescue JWT::DecodeError => e
+            jwt_not_authenticated(message: e.message) && return
```
mladenilic: Is trailing `return` call needed here?

mladenilic: `jwt_not_authenticated` is currently only called when JWT is not decoded properly. We should also handle cases when JWT is properly decoded, but the user record is missing.

lib/sorcery/controller/submodules/jwt_auth.rb
```diff
@@ -0,0 +1,85 @@
...
+          def jwt_decode(token)
+            HashWithIndifferentAccess.new(
+              JWT.decode(token, Config.jwt_secret_key)[0]
+            )
+          end
+
+          # Take token from header, by key defined in config
+          # With memoization
+          def jwt_from_header
+            @jwt_header_token ||= request.headers[Config.jwt_headers_key]
```
mladenilic: I think that more sensible default here would be to expect header value in one of the following formats:`Bearer <jwt-token>` or `Token <jwt-token>`. Same as what ActionController's authentication does – https://api.rubyonrails.org/classes/ActionController/HttpAuthentication/Token.html

Further more, does it makes sense to allow users to define how should token be extracted from request?

lib/sorcery/controller/submodules/jwt_auth.rb
```diff
@@ -0,0 +1,85 @@
...
+
+        module InstanceMethods
+          # This method return generated token if user can be authenticated
+          def jwt_auth(*credentials)
+            user = user_class.authenticate(*credentials)
+            if user
+              now = Time.current
+              default_payload = {
+                  sub: user.id,
+                  exp: (now + 3.days).to_i,
```
mladenilic: Token duration should be configurable.

athix:
Closing this out to reduce noise, see #239 and Sorcery/sorcery-rework#9 for the latest on JWT in Sorcery.

PR #173, opened 2019-01-31
allow BCrypt to have app-specific secret token

fursich:
Hello, thank you very much for the great work to maintain this great gem!

This PR introduces a small yet (imho) some important updates that hopefully helps Sorcery to enhance its password security.

## what it does

This patch adds an option that allows users to provide app-specific secret token (often called as 'pepper') with BCrypt, the default crypt-provider, which is currently missing (explained below).

To ensure compatibility, this change only takes place when the new option 'pepper' is provided - most importantly, this change **does NOT affect** on any of the gem's behavior, for those who have already set salt_join_token but leaves pepper option unset.

Tests are added to cover several BCrypt cases that have been missing, as well as this new feature.

## why this change is needed

Currently the config option 'salt_join_token' [has been ignored](https://github.com/Sorcery/sorcery/blob/88181d320388b7708dfdcc38efc265acd909f1aa/lib/sorcery/crypto_providers/bcrypt.rb#L43-L53) when BCrypt, the default crypto-provider is used, although [the comment](https://github.com/fursich/sorcery/blob/ebb483a31ca1d01d0ef61561617c9d5d625d76a0/lib/sorcery/model/config.rb#L16-L17) reads as if it works.

With sorcery, we have the password and salt on the same table (which would make sense itself), but due to this,  the two secret tokens can often be exposed altogether in case any accidental/malicious procedures take place.

Using BCrypt with decent costs (defaults 10) reduce the risk of brute forcing, but chances remain to have easy passwords (that users often choose) be guessed and broken when attackers obtain both passwords and the salts.

Allowing developers to set an app-specific secret token (that are usually kept separately from db) could enforce password security (or at least, give better chances to protect passwords) in case of potential leakage.

## other considerations

- devise [allows user to set `pepper` option](https://github.com/plataformatec/devise/blob/e3a00b27d19ba995891d7dd92394fe2900a789c2/lib/generators/templates/devise.rb#L116-L117), which basically is [a substring added at the end of the given token](https://github.com/plataformatec/devise/blob/715192a7709a4c02127afb067e66230061b82cf2/lib/devise/encryptor.rb#L8-L10).

- the leakage has common pattern named as [IDOR](https://www.owasp.org/index.php/Top_10_2007-Insecure_Direct_Object_Reference), so letting users to take measure on this type of incident would make a lot of sense. (unfortunately I personally have observed several incidents occurred with Rail apps, although I cannot name them..)

Let me know if there's anything to be clarified/updated.
Hope it makes sense!

### DIFF
--- a/lib/sorcery/crypto_providers/bcrypt.rb
+++ b/lib/sorcery/crypto_providers/bcrypt.rb
@@ -41,8 +41,8 @@ module Sorcery
     class BCrypt
       class << self
         # Setting the option :pepper allows users to append an app-specific secret token.
-        # basically equivalent to :salt_join_token option, but have different name to ensure
-        # backward compatibility in generating/matching passwards.
+        # Basically it's equivalent to :salt_join_token option, but have a different name to ensure
+        # backward compatibility in generating/matching passwords.
         attr_accessor :pepper
         # This is the :cost option for the BCrpyt library.
         # The higher the cost the more secure it is and the longer is take the generate a hash. By default this is 10.
--- a/lib/sorcery/crypto_providers/bcrypt.rb
+++ b/lib/sorcery/crypto_providers/bcrypt.rb
@@ -40,6 +40,10 @@ module Sorcery
     # You are good to go!
     class BCrypt
       class << self
+        # Setting the option :pepper allows users to append an app-specific secret token.
+        # basically equivalent to :salt_join_token option, but have different name to ensure
+        # backward compatibility in generating/matching passwards.
+        attr_accessor :pepper
         # This is the :cost option for the BCrpyt library.
         # The higher the cost the more secure it is and the longer is take the generate a hash. By default this is 10.
         # Set this to whatever you want, play around with it to get that perfect balance between
@@ -77,12 +81,13 @@ module Sorcery
 
         def reset!
           @cost = 10
+          @pepper = ''
         end
 
         private
 
         def join_tokens(tokens)
-          tokens.flatten.join
+          tokens.flatten.join.concat(pepper.to_s) # make sure to add pepper in case tokens have only one element
         end
 
         def new_from_hash(hash)
--- a/lib/sorcery/model.rb
+++ b/lib/sorcery/model.rb
@@ -142,6 +142,7 @@ module Sorcery
       def set_encryption_attributes
         @sorcery_config.encryption_provider.stretches = @sorcery_config.stretches if @sorcery_config.encryption_provider.respond_to?(:stretches) && @sorcery_config.stretches
         @sorcery_config.encryption_provider.join_token = @sorcery_config.salt_join_token if @sorcery_config.encryption_provider.respond_to?(:join_token) && @sorcery_config.salt_join_token
+        @sorcery_config.encryption_provider.pepper = @sorcery_config.pepper if @sorcery_config.encryption_provider.respond_to?(:pepper) && @sorcery_config.pepper
       end
 
       def add_config_inheritance
--- a/lib/sorcery/model/config.rb
+++ b/lib/sorcery/model/config.rb
@@ -12,11 +12,11 @@ module Sorcery
       attr_accessor :downcase_username_before_authenticating
       # change default crypted_password attribute.
       attr_accessor :crypted_password_attribute_name
-      # application-specific secret token that are joined with the password and its salt.
+      # application-specific secret token that is joined with the password and its salt.
       # Currently available with BCrypt (default crypt provider) only.
       attr_accessor :pepper
       # what pattern to use to join the password with the salt
-      # APPLICABLE TO MD5, SHA1, SHA256, SHA512. Other crypt providers (incl. BCrypt) ignores this parameter.
+      # APPLICABLE TO MD5, SHA1, SHA256, SHA512. Other crypt providers (incl. BCrypt) ignore this parameter.
       attr_accessor :salt_join_token
       # change default salt attribute.
       attr_accessor :salt_attribute_name
--- a/lib/sorcery/model/config.rb
+++ b/lib/sorcery/model/config.rb
@@ -12,7 +12,11 @@ module Sorcery
       attr_accessor :downcase_username_before_authenticating
       # change default crypted_password attribute.
       attr_accessor :crypted_password_attribute_name
+      # application-specific secret token that are joined with the password and its salt.
+      # Currently available with BCrypt (default crypt provider) only.
+      attr_accessor :pepper
       # what pattern to use to join the password with the salt
+      # APPLICABLE TO MD5, SHA1, SHA256, SHA512. Other crypt providers (incl. BCrypt) ignores this parameter.
       attr_accessor :salt_join_token
       # change default salt attribute.
       attr_accessor :salt_attribute_name
@@ -57,6 +61,7 @@ module Sorcery
           :@encryption_provider                  => CryptoProviders::BCrypt,
           :@custom_encryption_provider           => nil,
           :@encryption_key                       => nil,
+          :@pepper                               => '',
           :@salt_join_token                      => '',
           :@salt_attribute_name                  => :salt,
           :@stretches                            => nil,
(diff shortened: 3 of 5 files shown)


fursich:
(added minor cosmetic changes following Codacy's suggestions)

lib/sorcery/crypto_providers/bcrypt.rb
```diff
@@ -40,6 +40,10 @@ module CryptoProviders
     # You are good to go!
     class BCrypt
       class << self
+        # Setting the option :pepper allows users to append an app-specific secret token.
+        # basically equivalent to :salt_join_token option, but have different name to ensure
+        # backward compatibility in generating/matching passwards.
```
athix: Minor typo: `passwards` -> `passwords`

lib/sorcery/model.rb
```diff
@@ -142,6 +142,7 @@ def authentication_response(options = {})
       def set_encryption_attributes
         @sorcery_config.encryption_provider.stretches = @sorcery_config.stretches if @sorcery_config.encryption_provider.respond_to?(:stretches) && @sorcery_config.stretches
         @sorcery_config.encryption_provider.join_token = @sorcery_config.salt_join_token if @sorcery_config.encryption_provider.respond_to?(:join_token) && @sorcery_config.salt_join_token
+        @sorcery_config.encryption_provider.pepper = @sorcery_config.pepper if @sorcery_config.encryption_provider.respond_to?(:pepper) && @sorcery_config.pepper
```
athix: Should the conditional check `sorcery_config.pepper.present?` here? I believe ruby treats an empty string as a truthy object.

lib/sorcery/model/config.rb
```diff
@@ -12,7 +12,11 @@ class Config
       attr_accessor :downcase_username_before_authenticating
       # change default crypted_password attribute.
       attr_accessor :crypted_password_attribute_name
+      # application-specific secret token that are joined with the password and its salt.
```
athix: Minor typo: `that are joined` -> `that is joined`

spec/shared_examples/user_shared_examples.rb
```diff
@@ -488,6 +503,32 @@ def self.matches?(crypted, *tokens)
 
       expect(user.crypted_password).to eq Sorcery::CryptoProviders::SHA512.encrypt('secret', user.salt)
     end
+
+    context '' do
```
athix: Using an empty context feels like a bad practice. Can we update this to reflect why it's being used, or remove it?

lib/sorcery/crypto_providers/bcrypt.rb
fursich: 😅

fursich:
@athix 
Updated the patch based on your suggestions - let me know if there's anything left to be done.
I'm happy to squash the commits if necessary (upon your approval).

Thanks again for your amazingly swift feedback :)

athix:
@fursich We use squash merge, so no need to worry about squashing locally. Still uncertain on the `.present?` issue, so if you could just address that. Thanks!

lib/sorcery/model.rb
fursich: Adding an empty `'pepper'` does nothing actually - when appending an empty string at the end of the token, the original token remains as it is. (shall I add test cases for that maybe?)

Also providing empty string explicitly has an effect to [remove the prior pepper](https://github.com/Sorcery/sorcery/blob/5746cec165998fd377769515d7e8ea4cc4539938/spec/shared_examples/user_shared_examples.rb#L522) (usually should not be changed, but would be an expected behavior as we see with `salt_join_token` ). If we ignore empty string, there is no way to remove the pepper once set.

I could add an check here though, if you think it makes better senses.
In that case, the only case we need to clarify is how to treat an blank string (i.e. string with spaces)

```ruby
pry(main)> puts 'true' if "\s"
(pry):101: warning: string literal in condition
true
=> nil
pry(main)> puts 'true' if "\s".present?
=> nil
```

So if we use `present?`, with space-filled tokens, it adds no peppers, while with the original approach, it actually appends spaces at the end of the string (they ends up with different thing as digest)

```ruby
pry(main)> x = BCrypt::Password.create("password\s")
=> "$2a$10$Q4ONgGV.uwtJ1iQX/Sr35u4wlDLZ0A28CKCG2s3nbRYHR1tRxg8rS"
pry(main)> x == "password"
=> false
```

Although it looks like a corner case, I thought it might be better to align this behavior with the original `salt_join_token`.

Having said that, of course, I would be happy to follow your call. What would you think?

fursich:
@athix  thanks again, just updated my answer :)
Long story in short, I'm a little worried about the corner cases, as it will behave differently from `salt_join_token`.
Anyway let me know how you think, I'm happy to update them if needed:)

lib/sorcery/model.rb
athix: If there's tests that ensure it doesn't break backwards compatibility, I think that would be ideal. Beyond that, I'll have to dig into this a little deeper to get a better understanding.

athix:
@Ch4s3 if you have the time, can you review this as well? It covers some important changes to how Sorcery handles the bcrypt password hashing.

spec/sorcery_crypto_providers_spec.rb
```diff
@@ -207,6 +207,36 @@
...
+      end
+    end
+
+    context "when pepper is an empty string (default)" do
+      before(:each) do
+        Sorcery::CryptoProviders::BCrypt.pepper = ''
+        @digest = Sorcery::CryptoProviders::BCrypt.encrypt(@tokens) # a BCrypt::Password object
+      end
+
+      # make sure the default pepper '' does nothing
```
fursich: Added this line to make double sure that the encryption behavior shouldn't change before/after setting this option.

spec/shared_examples/user_shared_examples.rb
```diff
@@ -527,6 +527,30 @@ def self.matches?(crypted, *tokens)
 
       expect(user.crypted_password).to eq Sorcery::CryptoProviders::BCrypt.encrypt('secret', user.salt)
     end
+
+    it 'if pepper is empty string (default) does not use pepper to encrypt' do
```
fursich: this case will fail if we add `present?` to [#set_encryption_attributes method](https://github.com/Sorcery/sorcery/blob/ba659b9a8e7564ea6a202a1419445ade3a02697a/lib/sorcery/model.rb#L145) (discussed earlier), as `present?` returns false when `''` is specified, and thus keeps previously-set pepper unchanged.

spec/sorcery_crypto_providers_spec.rb
```diff
@@ -207,6 +207,36 @@
...
+      it 'matches token encrypted with salt only (without pepper)' do
+        expect(@digest).to eq @tokens.flatten.join # keep consistency with the older versions of #join_token
```
fursich: the right hand side is actually current `#join_token` (versions without pepper).
This case is here to check the new token generated with default pepper (i.e. empty string) stays the same as the token that would have been generated by the current Sorcery.

fursich:
To answer your point (which I found a good point to care), I added a few test cases above.

The main purpose here is to make sure the token generated with default pepper stays unchanged with the token that would have been generated by the versions before this pepper introduced. (meaning, it keeps working without breakage unless you explicitly set a new pepper)

Let me know if I could help any further. Hope that makes senes! :)

lib/sorcery/model.rb
fursich: Just added a couple of test cases to describe how this new feature behaves, especially to double check that no breakage occurs with the update.
Thanks for your input, will wait for your feedback/questions if any.

athix:
Thanks @fursich! I did a quick once over, and everything is looking good so far. I'll give this a really thorough look-over when I have more time available, seeing as how critical the password hashing process is.

fursich:
Hi @athix ,
just a friendly reminder, but did you have a chance to take look on it?

I"m guessing, it probably makes sense to remove the slightly misleading explanation around `salt_join_token`, since it is being totally ignored by default BCrypt module currently.

This patch basically gives its working alternatives, but under a different name (to make sure no breakage occurs when `salt_join_token` had been previously set)

Anyway let me know if there is anything to work on (no pressure please!) 😸

athix:
@fursich thanks for the reminder! I'll set aside some time this weekend.

fursich:
Hi @athix 
Thanks really a lot for putting your weekend time on this!
Glad to hear it gets merged 😸

PR #175, opened 2019-02-19
Add failure reason argument for `after_failed_login` callback

pinewong:
Follow https://github.com/Sorcery/sorcery/pull/41, maybe we need the failure reason in callbacks

### DIFF
--- a/lib/sorcery/controller/submodules/brute_force_protection.rb
+++ b/lib/sorcery/controller/submodules/brute_force_protection.rb
@@ -20,7 +20,7 @@ module Sorcery
 
           # Increments the failed logins counter on every failed login.
           # Runs as a hook after a failed login.
-          def update_failed_logins_count!(credentials)
+          def update_failed_logins_count!(credentials, _failure_reason)
             user = user_class.sorcery_adapter.find_by_credentials(credentials)
             user.register_failed_login! if user
           end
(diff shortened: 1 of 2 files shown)


lib/sorcery/controller/submodules/brute_force_protection.rb
```diff
@@ -20,7 +20,7 @@ module InstanceMethods
 
           # Increments the failed logins counter on every failed login.
           # Runs as a hook after a failed login.
-          def update_failed_logins_count!(credentials)
+          def update_failed_logins_count!(credentials, _failure_reason)
```
athix: This appears to be the only callback that was changed, and failure_reason is unused. @pinewong can you elaborate on this change?

athix (dismissed):

lib/sorcery/controller/submodules/brute_force_protection.rb
pinewong: `failure_reason` is useful for `after_failed_login` callback, so I added it. Then I searched the current project, and found there was one place(we are referring to) use it, I have to modify it synchronously.

athix: Ah, yeah that makes sense.

athix:
@songhuangcn The failure_reason is being expanded to be `:status` due to the addition of MFA in the near future, but I think it's pretty reasonable to pass the status to any login-related callbacks. I'll get this added to v1.

PR #179, opened 2019-03-19
Make the `remember_me_token` cookie name configurable

robinvdvleuten:
This PR makes the default `remember_me_token` cookie name configurable to whatever seems to fit for the application.


lib/generators/sorcery/templates/initializer.rb
```diff
@@ -26,6 +26,11 @@
   #
   # config.cookie_domain =
 
+  # Set name of the remember_me cookie
+  # Default: `remember_me_token`
```
athix: This should probably be `:remember_me_token` so it's clear that it's a symbol (and not say, a String) here.

lib/sorcery/controller/submodules/remember_me.rb
```diff
@@ -36,13 +39,13 @@ def remember_me!
           # Clears the cookie, and depending on the value of remember_me_token_persist_globally, may clear the token value.
           def forget_me!
             current_user.forget_me!
-            cookies.delete(:remember_me_token, domain: Config.cookie_domain)
+            cookies.delete(Config.remember_me_cookie_name, domain: Config.cookie_domain)
```
athix: It may make sense to call `.to_sym` here so that users can use a string or symbol in the config, similar to our other config values.

lib/sorcery/controller/submodules/remember_me.rb
```diff
@@ -36,13 +39,13 @@ def remember_me!
...
           end
 
           # Clears the cookie, and clears the token value.
           def force_forget_me!
             current_user.force_forget_me!
-            cookies.delete(:remember_me_token, domain: Config.cookie_domain)
+            cookies.delete(Config.remember_me_cookie_name, domain: Config.cookie_domain)
```
athix: Similarly, these other instances would also be updated to call `.to_sym`

spec/controllers/controller_remember_me_spec.rb
```diff
@@ -38,7 +52,7 @@
       pending 'Test previously broken, functionality might not be working here.'
       expect(cookies['remember_me_token']).to be_nil
     end
-
+
```
athix: Looks like some trailing whitespace was created here, please remove.

spec/controllers/controller_remember_me_spec.rb
```diff
@@ -58,6 +72,22 @@
       expect(cookies.signed['remember_me_token']).to eq assigns[:user].remember_me_token
     end
 
+    it 'login(email,password,remember_me) logs user in and remembers with custom cookie name' do
+      sorcery_controller_property_set(:remember_me_cookie_name, :custom)
+
+      expect(User).to receive(:authenticate).with('bla@bla.com', 'secret', '1') { |&block| block.call(user, nil) }
+      expect(user).to receive(:remember_me!)
+      expect(user).to receive(:remember_me_token).and_return('abracadabra').twice
```
athix: Why does this receive the token twice? I'm not very familiar with the callback system, but this looks like it might be a side-effect of the double callback bug in `0.13.0`

robinvdvleuten:
@athix I've implemented your remarks. Could you take a look again?

spec/controllers/controller_remember_me_spec.rb
robinvdvleuten: @athix to be honest, I copied it over from the testcase above this one. So it could be part of the double callback bug. When I remove the `twice` method though, it gives an error. Could it be that the double callback bug is still relevant?

kentarohorie:
@robinvdvleuten It seems that CI failed.

PR #180, opened 2019-03-19
Add missing remember_me attributes to config

robinvdvleuten:
The `remember_me` attributes were missing in the initialized configuration.

### DIFF
--- a/lib/generators/sorcery/templates/initializer.rb
+++ b/lib/generators/sorcery/templates/initializer.rb
@@ -281,6 +281,16 @@ Rails.application.config.sorcery.configure do |config|
     # user.subclasses_inherit_config =
 
     # -- remember_me --
+    # change default remember_me_token attribute.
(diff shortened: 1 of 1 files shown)


lib/generators/sorcery/templates/initializer.rb
```diff
@@ -281,6 +281,16 @@
     # user.subclasses_inherit_config =
 
     # -- remember_me --
+    # change default remember_me_token attribute.
```
athix: Looks like this was accidentally included from your code in the other PR.

robinvdvleuten: It is the column name of the `remember_me_token` attribute, not the name of the remember me cookie.

robinvdvleuten:
@athix it looks like you confused the `remember_me_token` column attribute with the cookie name in my other PR. Could you take a look again?

athix (approved):

PR #183, opened 2019-04-25
Fix typo in the base providers :auth_hash

andrewhood125:
Based on the :respond_to? it looks like this was intended to be storing
the :expires_in method.


athix:
This appears to have been the case for at least 5 years, so I would guess that this was either intentional, or gets overwritten by each individual provider and does not matter regardless. If you still think this is potentially an issue, please let me know and we can reopen this.

PR #184, opened 2019-04-25
Add limitless counting failed login function

hairgai:
Hi.

This function is adding limitless counting 'failed login'.
I think, if stopped counting 'failed login' after reached limit, we can't tell the reason of lock either user's action or brute force attack, so I added.

If you don't need, please close this PR
Thanks.

### DIFF
--- a/lib/sorcery/model/submodules/brute_force_protection.rb
+++ b/lib/sorcery/model/submodules/brute_force_protection.rb
@@ -14,6 +14,8 @@ module Sorcery
                           :consecutive_login_retries_amount_limit,    # how many failed logins allowed.
                           :login_lock_time_period,                    # how long the user should be banned.
                           # in seconds. 0 for permanent.
+                          :limitless_counting_failed_login,           # When true, continue to count failed login
+                          # after reached limit.
                           :unlock_token_attribute_name,               # Unlock token attribute name
                           :unlock_token_email_method_name,            # Mailer method name
                           :unlock_token_mailer_disabled,              # When true, dont send unlock token via email
@@ -25,6 +27,7 @@ module Sorcery
                              :@lock_expires_at_attribute_name                  => :lock_expires_at,
                              :@consecutive_login_retries_amount_limit          => 50,
                              :@login_lock_time_period                          => 60 * 60,
+                             :@limitless_counting_failed_login                 => false,
 
                              :@unlock_token_attribute_name                     => :unlock_token,
                              :@unlock_token_email_method_name                  => :send_unlock_token_email,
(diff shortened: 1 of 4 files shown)


lib/sorcery/model/submodules/brute_force_protection.rb
```diff
@@ -63,13 +66,14 @@ def define_brute_force_protection_fields
         module InstanceMethods
           # Called by the controller to increment the failed logins counter.
           # Calls 'login_lock!' if login retries limit was reached.
-          def register_failed_login!
+          def register_failed_login!(password)
```
mladenilic: Is the password parameter necessary here? The method simply registers a failed login attempt, its behavior should not depend on the provided password. 

Can we just reverse the order of increment call and `return unless login_unlocked?`. Will that have the desired effect?

lib/sorcery/model/submodules/brute_force_protection.rb
```diff
@@ -14,6 +14,8 @@ def self.included(base)
                           :consecutive_login_retries_amount_limit,    # how many failed logins allowed.
                           :login_lock_time_period,                    # how long the user should be banned.
                           # in seconds. 0 for permanent.
+                          :limitless_counting_failed_login,           # When true, continue to count failed login
```
mladenilic: I don't think another config option is needed here. Is there a reason why someone would want to disable this option?

lib/sorcery/model/submodules/brute_force_protection.rb
```diff
@@ -63,13 +66,14 @@ def define_brute_force_protection_fields
...
-          def register_failed_login!
+          def register_failed_login!(password)
             config = sorcery_config
-            return unless login_unlocked?
 
-            sorcery_adapter.increment(config.failed_logins_count_attribute_name)
+            return if login_locked? && !config.limitless_counting_failed_login
 
-            return unless send(config.failed_logins_count_attribute_name) >= config.consecutive_login_retries_amount_limit
+            sorcery_adapter.increment(config.failed_logins_count_attribute_name) unless valid_password?(password)
```
athix, line 74: I suspect this might open the window to timing attacks, especially considering the trip time to a DB may be significant in certain setups.

lib/sorcery/model/submodules/brute_force_protection.rb
athix: As mentioned, passing the password to this callback is unnecessary at best, and malicious at worst.

athix (changes requested):
I'll take another look at this later, some significant concerns with this particular implementation however.

PR #190, opened 2019-05-22
Allow for custom providers with multi-word class names.

mvandenbeuken:
Currently sorcery will not handle custom providers with names like: ExampleProvider. config.example_provider would search for a class called Example_provider rather than ExampleProvider. Using .classify rather than .capitalize will fix this.

I've added an example provider implementation under the spec/support directory, and have a test that ensures that it is loaded properly via config. The existing tests for existing providers are also coming back with successes.

### DIFF
--- a/lib/sorcery/version.rb
+++ b/lib/sorcery/version.rb
@@ -1,3 +1,3 @@
 module Sorcery
-  VERSION = '0.14.0'.freeze
+  VERSION = '0.15.0'.freeze
 end
--- a/lib/sorcery/version.rb
+++ b/lib/sorcery/version.rb
@@ -1,3 +1,3 @@
 module Sorcery
-  VERSION = '0.15.0'.freeze
+  VERSION = '0.14.0'.freeze
 end
--- a/lib/sorcery/controller/submodules/external.rb
+++ b/lib/sorcery/controller/submodules/external.rb
@@ -38,7 +38,7 @@ module Sorcery
                 providers.each do |name|
                   class_eval <<-RUBY, __FILE__, __LINE__ + 1
                     def self.#{name}
-                      @#{name} ||= Sorcery::Providers.const_get('#{name}'.to_s.capitalize).new
+                      @#{name} ||= Sorcery::Providers.const_get('#{name}'.to_s.classify).new
                     end
                   RUBY
                 end
--- a/spec/providers/example_provider_spec.rb
+++ b/spec/providers/example_provider_spec.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'sorcery/providers/base'
+
+describe Sorcery::Providers::ExampleProvider do
+  before(:all) do
+    sorcery_reload!([:external])
+    sorcery_controller_property_set(:external_providers, [:example_provider])
+  end
+
+  context 'fetching a multi-word custom provider' do
+    it 'returns the provider' do
+      expect(Sorcery::Controller::Config.example_provider).to be_a(Sorcery::Providers::ExampleProvider)
+    end
+  end
+end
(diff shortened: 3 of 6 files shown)


lib/sorcery/version.rb
```diff
@@ -1,3 +1,3 @@
 module Sorcery
-  VERSION = '0.14.0'.freeze
+  VERSION = '0.15.0'.freeze
```
chhga: not sure if we want to change version.

mvandenbeuken:
@chhga Thanks! I've added a spec to cover single-word custom providers. I've also reverted the version back to 0.14.0.

athix:
@chhga Yup, looks good to me as well. Merging.

murtali:
curious @mvandenbeuken did you add a custom provider without forking the gem and adding it into the submodule folder?

athix:
@murtali I'd imagine you should be able to require your custom provider, and as long as it's conformant to the standard the rest of the providers follow, and you add it to the providers list in your config, it should work.

murtali:
@athix  -- hmm I tried doing the following:

```
# added file in {Rails.root}/lib/my_custom_provider.rb
```

```
module Sorcery
  module Providers
    class MyCustomProvider < Base

      def initialize
        super
      end
      
      def login_url(_params, _session)
      end

      def get_user_hash(access_token)
      end

      def process_callback(params, _session)
      end
    end
  end
end
```

and then in `initializers/sorcery.rb` i added the following:
```
  # -- external --
  # What providers are supported by this app
  # i.e. [:twitter, :facebook, :github, :linkedin, :xing, :google, :liveid, :salesforce, :slack, :line].
  # Default: `[]`
  #
  config.external_providers = [:my_custom_provider]
  config.my_custom_provider.callback_url = "https://my_url.com/callback"
```

and it's throwing the following error when I tried starting the rails server:
```
3: from /Users/me/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/sorcery-0.16.1/lib/sorcery/controller/config.rb:61:in `configure!'
2: from /Users/me/my_app/config/initializers/sorcery.rb:87:in `block in <main>'
1: from /Users/me/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/sorcery-

0.16.1/lib/sorcery/controller/submodules/external.rb:43:in `my_custom_provider'
/Users/me/.rbenv/versions/2.6.4/lib/ruby/gems/2.6.0/gems/sorcery-0.16.1/lib/sorcery/controller/submodules/external.rb:43:in `const_get': uninitialized constant Sorcery::Providers::MyCustomProvider (NameError)
```

it seems to be looking for submodule in the gem source?

Should I be doing this in another way?

(thanks again for the response here)

athix:
@murtali You need to tell the application where it can find that class: `uninitialized constant Sorcery::Providers::MyCustomProvider (NameError)`

Currently Sorcery tries to autoload that constant, but fails to find the class because neither your application nor Sorcery provides it.

Within `config/application.rb`:

```ruby
config.autoload_paths += %W(#{config.root}/lib)
```

As for getting it to auto discover the class within lib, I'm honestly not sure what the best approach would be. It's been a while since I've added a class to an already existing/loaded module. I don't know if it might get confused because the Sorcery::Providers module already has been loaded...What may work if Ruby works the way I would hope it would in this scenario, is:

```ruby
# lib/sorcery.rb
module Sorcery
  autoload :Providers, 'sorcery/providers'
end
```

```ruby
# lib/sorcery/providers.rb
module Sorcery
  module Providers
    autoload :MyCustomProvider, 'sorcery/providers/my_custom_provider'
  end
end
```

```ruby
# lib/sorcery/providers/my_custom_provider.rb
class Sorcery::Providers::MyCustomProvider < Sorcery::Providers::Base
  # [...]
end
```

PR #191, opened 2019-05-28
Load user email in LinkedIn provider

JakubSzajna:
As per https://github.com/Sorcery/sorcery/blob/master/lib/generators/sorcery/templates/migration/core.rb user email is a mandatory field in the DB. Thus, the recent fix for LinkedIn provider will produce an error when you try to register a new profile with it.

This fix loads user email from the additional endpoint in LinkedIn API v2 following this description: https://docs.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/sign-in-with-linkedin?context=linkedin/consumer/context#retrieving-member-email-address

### DIFF
--- a/lib/generators/sorcery/templates/initializer.rb
+++ b/lib/generators/sorcery/templates/initializer.rb
@@ -88,6 +88,9 @@ Rails.application.config.sorcery.configure do |config|
   #
   # config.ca_file =
 
+  # Linkedin requires r_emailaddress scope to fetch user's email address.
+  # You can skip it if email field is not mandatory in your database (by default it is).
+  #
   # config.linkedin.key = ""
   # config.linkedin.secret = ""
   # config.linkedin.callback_url = "http://0.0.0.0:3000/oauth/callback?provider=linkedin"
--- a/lib/generators/sorcery/templates/initializer.rb
+++ b/lib/generators/sorcery/templates/initializer.rb
@@ -89,7 +89,8 @@ Rails.application.config.sorcery.configure do |config|
   # config.ca_file =
 
   # Linkedin requires r_emailaddress scope to fetch user's email address.
-  # You can skip it if email field is not mandatory in your database (by default it is).
+  # You can skip including the email field if you use an intermediary signup form. (using build_from method).
+  # The r_emailaddress scope is only necessary if you are using the create_from method directly.
   #
   # config.linkedin.key = ""
   # config.linkedin.secret = ""
--- a/lib/generators/sorcery/templates/initializer.rb
+++ b/lib/generators/sorcery/templates/initializer.rb
@@ -91,8 +91,12 @@ Rails.application.config.sorcery.configure do |config|
   # config.linkedin.key = ""
   # config.linkedin.secret = ""
   # config.linkedin.callback_url = "http://0.0.0.0:3000/oauth/callback?provider=linkedin"
-  # config.linkedin.user_info_mapping = {first_name: "firstName", last_name: "lastName"}
-  # config.linkedin.scope = "r_basicprofile"
+  # config.linkedin.user_info_mapping = {
+  #   first_name: 'localizedFirstName',
+  #   last_name:  'localizedLastName',
+  #   email:      'emailAddress'
+  # }
+  # config.linkedin.scope = "r_liteprofile r_emailaddress"
   #
   #
   # For information about XING API:
--- a/lib/sorcery/providers/linkedin.rb
+++ b/lib/sorcery/providers/linkedin.rb
@@ -35,7 +35,6 @@ module Sorcery
       # calculates and returns the url to which the user should be redirected,
       # to get authenticated at the external provider's site.
       def login_url(_params, _session)
-        force_email_scope
         authorize_url(authorize_url: auth_url)
       end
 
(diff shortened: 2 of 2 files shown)


chhga:
Do we have enough specs to test this case?

lib/generators/sorcery/templates/initializer.rb
```diff
@@ -91,8 +91,12 @@
   # config.linkedin.key = ""
   # config.linkedin.secret = ""
   # config.linkedin.callback_url = "http://0.0.0.0:3000/oauth/callback?provider=linkedin"
-  # config.linkedin.user_info_mapping = {first_name: "firstName", last_name: "lastName"}
-  # config.linkedin.scope = "r_basicprofile"
+  # config.linkedin.user_info_mapping = {
+  #   first_name: 'localizedFirstName',
```
athix: Looking at the [docs](https://developer.linkedin.com/docs/fields/basic-profile), it looks like the fields were updated to `first-name` and `last-name`. I do remember these being nested hashes, so I wouldn't be surprised if `localizedXName` is what you're supposed to use. @JakubSzajna can you link to the documentation or otherwise confirm that this has been tested?

lib/sorcery/providers/linkedin.rb
```diff
@@ -9,25 +9,26 @@ module Providers
     class Linkedin < Base
       include Protocols::Oauth2
 
-      attr_accessor :auth_url, :scope, :token_url, :user_info_url
+      attr_accessor :auth_url, :scope, :token_url, :user_info_url, :email_info_url
```
athix: Do we actually need this? [The docs](https://developer.linkedin.com/docs/fields/basic-profile) seem to indicate all you need is to include `r_emailaddress` in the scope.

lib/generators/sorcery/templates/initializer.rb
JakubSzajna: Hey @athix, I was not able to find it anywhere in the docs, unfortunately, so I'm a little bit worried it might stop working at some point. But the code here is taken from my current working configuration, so yes, I've tested this and it works fine.

athix: Sounds good, it's broken as-is so we can go ahead and update it until LinkedIn decides to break it again. :shipit:

athix:
@chhga unfortunately we don't have reliable specs for any of the OAuth providers. I don't know that it's worth the dev-time to develop it however, as the plan is to switch to Omniauth at some point. That being said, I wouldn't be opposed if someone wanted to put the time into creating specs. We can definitely use more coverage.

lib/sorcery/providers/linkedin.rb
```diff
@@ -45,6 +46,15 @@ def process_callback(params, _session)
 
         get_access_token(args, token_url: token_url, token_method: :post)
       end
+
+      def get_user_info(access_token)
+        response       = access_token.get(user_info_url)
+        email_response = access_token.get(email_info_url)
```
athix: I'm concerned about this breaking when an application isn't trying to use the `r_emailaddress` scope. @JakubSzajna can you test this while passing only `r_liteprofile` to ensure it still works?

JakubSzajna: You are right @athix 
This breaks the application as there is an `OAuth2::Error` exception.
I've forced `r_emailaddress` to exist in `@scope` as skipping the email request would get us back to the beginning.
I've tested it and it worked in my app.

athix: Forcing a scope is not ideal, imo. Instead, I think we should check the scope for `r_emailaddress` and conditionally do the email_response call / merge.

```ruby
def scope_includes_r_emailaddress?
  scope = @scope.split(' ')
  scope.include?('r_emailaddress')
end
```

```ruby
if scope_includes_r_emailaddress?
  email_info = JSON.parse([...])

  user_info.merge([...])
end

user_info
```

JakubSzajna: Ah, I've missed your solution and did it not exactly the same way, but it works 👍

lib/generators/sorcery/templates/initializer.rb
```diff
@@ -88,11 +88,18 @@
   #
   # config.ca_file =
 
+  # Linkedin requires r_emailaddress scope to fetch user's email address.
+  # You can skip it if email field is not mandatory in your database (by default it is).
```
athix: You can also skip including the email field if you use an intermediary signup form. (using `build_from` method)

The `r_emailaddress` scope is only necessary if you are using the `create_from` method directly.

lib/sorcery/providers/linkedin.rb
```diff
@@ -45,6 +46,30 @@ def process_callback(params, _session)
...
       end
+
+      def get_user_info(access_token)
+        response  = access_token.get(user_info_url)
+        user_info = JSON.parse(response.body)
+
+        if email_in_scope?
+          email = fetch_email(access_token)
+
+          return user_info.merge(email)
```
athix: Nice, this is more elegant. 👍

lib/sorcery/providers/linkedin.rb
```diff
@@ -45,6 +46,30 @@ def process_callback(params, _session)
...
+        end
+
+        user_info
+      end
+
+      def email_in_scope?
+        scope.include?('r_emailaddress')
```
athix: I forgot you could call include directly on the string. 👍

lib/generators/sorcery/templates/initializer.rb
JakubSzajna: Adjusted 👍

athix: Thanks!

athix:
Thank you @JakubSzajna for your work on this! Merging.

PR #196, opened 2019-06-21
Account lock is released even if password is wrong if login lock has …

return-nil:
Hi.

If an incorrect password is entered after the account lock has expired, the account lock will not be released.
I thought it would be better to unlock the account even if the wrong password was entered

If you don't need, please close this PR
Thanks.

### DIFF
--- a/lib/sorcery/model.rb
+++ b/lib/sorcery/model.rb
@@ -102,20 +102,20 @@ module Sorcery
 
         set_encryption_attributes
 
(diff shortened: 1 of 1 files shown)


athix:
Keeping an account locked until the correct password is used sounds like intended behavior, please reopen/comment if that's not the case.

return-nil:
I thought it would be better to unlock if the lock has expired.

PR #199, opened 2019-07-16
Use id instead of uid for VK provider

anaumov:
VK has changed the API and it returns `user_info` in such way
```
{
  email: 'someone@vk.com',
  id: 12300321
}
```
Therefore `user_hash[:user_info]['uid']` is `nil` and sorcery does not create authentication for user. This patch should fix this. 

Not sure how to add spec for this case. Let me know if you have any questions.


anaumov:
@athix can you take a look?

athix:
I'm currently slammed, but I'll try to get caught up with Sorcery as soon as possible.

Issue #125, opened 2018-05-16
Faraday::SSLError


bragovo:
On every Ubuntu systems with rbenv and ruby 2.5.0 not working auth through VK provider. On Mac everything fine and FB provider work fine too. Is that problem with sorcery or faraday or Ubuntu?

bragovo:
Set `config.ca_file = '/usr/lib/ssl/certs/ca-certificates.crt'` and now auth fine. Maybe default ca_file is outdated?

Ch4s3:
@bragovo could be? Sorry I don't know too much about this and I've never used the FB provider. could you check with the Faraday folks?

athix:
The pr #200 addresses this, but is blocked by validation that the new CA is correct / safe.

athix:
The ideal resolution for the whole CA issue will be offloading that responsibility to Omniauth, imo.

Assuming there are no issues switching to Omniauth, this will be fixed in v1.

PR #200, opened 2019-07-25
Update sorcery CA bundle

anaumov:
I faced with SLL issue on VK auth. An issue described here https://github.com/Sorcery/sorcery/issues/125. I found that CA bundle was updated 6 years ago. Here a new one. I downloaded it from https://curl.haxx.se/docs/caextract.html. Let me know if you have any questions.


anaumov:
@athix can you take a look?

athix:
Updating the certificate authorities sounds like a high-risk change, so I'm hesitant to merge this without thoroughly validating the new list. If you have resources to help validate the authenticity, I can get this in sooner. Otherwise it will have to wait until I have time to look further into the issue.

Sorry for the inconvenience! In the short-term, you should be able to point your project at your fork with the ca changes.

anaumov:
@athix Not sure how to validate it automatically. Will take a look and get back.

athix:
Hi @anaumov, any luck?

anaumov:
@athix will get back to this PR next week. Sorry for the delay.

anaumov:
I did research about this PEM file. It made by the tool [mk-ca-bundle](https://curl.haxx.se/docs/mk-ca-bundle.html) from creators of the curl. It downloads a file with all authorized providers from Mozilla. Then it converts a file to PEM format. That's it.

I looks ok to me to use the tool from creators of curl and Mozilla CA data. What do you think, @athix?

Resources:
[Mozzila certificate store](https://www.mozilla.org/en-US/about/governance/policies/security-group/certs/)
[Mozzila certdata file](http://mxr.mozilla.org/mozilla-central/source/security/nss/lib/ckfw/builtins/certdata.txt)
[mk-ca-bundle man page](https://curl.haxx.se/docs/mk-ca-bundle.html)
[mk-ca-bundle source code](https://github.com/curl/curl/blob/master/lib/mk-ca-bundle.pl)

athix:
@anaumov I'll look into implementing this in the v1 rework, although it may become obsolete if we move the provider system to use omniauth instead of directly calling oauth2.

PR #202, opened 2019-08-04
enable to change login session key

abgata20000:
I wanna change to session key

Please accept pull request or enable to change session key it in other ways


mladenilic:
Does it make more sense to move the session key to initializer and use `Config.user_id_session_key` or something along those lines?

This way one can simply change the key in the initializer.

Just out of curiosity, what's the reason for changing the session key?

abgata20000:
@mladenilic 
I agree with your opinion
It is only because my intention was likely to be easily changed
Using Config.user_id_session_key is more reasonable

I will try to adjust again using the method of Config.user_id_session_key

athix:
#250 addresses this (configurable session_key)

PR #209, opened 2019-11-10
Fix deprecation warnings in Rails 6

tobyaherbert:
Sorcery currently uses ```defined?``` checks in ```engine.rb``` which causes Rails to autoload those constants during initialization. This generates deprecation warnings when using Rails 6:

```
DEPRECATION WARNING: Initialization autoloaded the constants ActionText::ContentHelper and ActionText::TagHelper.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed during
initialization does not run again. So, if you reload ActionText::ContentHelper, for example,
the expected changes won't be reflected in that stale Module object.

These autoloaded constants have been unloaded.

Please, check the "Autoloading and Reloading Constants" guide for solutions.
 (called from <top (required)> at /path/to/rails-6-app/config/environment.rb:5)
```

Using the ```#on_load``` method provided ActiveSupport fixes this problem.

---

I found these similar GitHub Issues whilst researching this:
- https://github.com/rollbar/rollbar-gem/issues/880
- https://github.com/rails/rails/issues/36546

PR #212, opened 2019-11-27
Remove deprecation warning in Rails 6

multiplegeorges:
Hi!

First off, great gem. It's my go to for auth and user management.

Description
--

This PR removes the following deprecation warning in Rails 6:

```
DEPRECATION WARNING: Initialization autoloaded the constant ActionText::ContentHelper.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed during
initialization does not run again. So, if you reload ActionText::ContentHelper, for example,
the expected changes won't be reflected in that stale Module object.

This autoloaded constant has been unloaded.

Please, check the "Autoloading and Reloading Constants" guide for solutions.
```

Testing
--
I've manually tested this on my current project and it works.

I'm not sure of what actual tests to add so verify this with all the versions of Sorcery and supported Rails versions.

### DIFF
--- a/lib/sorcery/engine.rb
+++ b/lib/sorcery/engine.rb
@@ -9,11 +9,11 @@ module Sorcery
 
     initializer 'extend Controller with sorcery' do
       # TODO: Should this include a modified version of the helper methods?
-      if defined?(ActionController::API)
+      ActiveSupport.on_load(:action_controller_api) do
         ActionController::API.send(:include, Sorcery::Controller)
       end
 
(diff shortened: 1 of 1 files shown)


mladenilic:
Hi there. This looks like a duplicate of #209.

multiplegeorges:
Indeed it is! Didn't notice that, they look identical.

PR #214, opened 2019-11-27
Remove MySQL database creation call

### LINKED ISSUE #213: TravisCI Builds Failing
It appears the configuration for the test database creation needs to be updated for TravisCI. All builds are failing due to the following error:

```
$ mysql -e 'create database sorcery_test;'

ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock' (2)

The command "mysql -e 'create database sorcery_test;'" failed and exited with 1 during .
```

athix:
Attempting to fix #213


mladenilic:
Build now fails cause of missing manifest file, I think this PR solves that issue #211.

PR #216, opened 2019-11-29
Add forget_me! and force_forget_me! test cases

mladenilic:
This PR completes two test cases that were marked as `pending`.


mladenilic:
@athix If you get a chance, take a look at this. I would like to include this in next release as well.

PR just re-enables previously disabled test cases. No actual behaviour is changed.

athix:
Thanks for adding more specs! :)

PR #219, opened 2019-12-24
Fix typo

mingrammer:


spec/controllers/controller_spec.rb
```diff
@@ -171,7 +171,7 @@
     # --- auto_login(user) ---
     specify { should respond_to(:auto_login) }
 
-    it 'auto_login(user) los in a user instance' do
+    it 'auto_login(user) logs in an user instance' do
```
mladenilic: Shouldn't it be `a user instance` instead of `an user instance`? :)

mingrammer: @mladenilic I fixed that! Thank you

Issue #231, opened 2020-04-30
brute force vulnerability


futuretap:
Sorcery 0.14.0 is vulnerable to brute force attacks even if a login_lock_time_period is configured.

If an attacker continues their login attempts after the initial login_lock_time_period has passed, Sorcery no longer rejects those login attempts until eventually the attacker has guessed the right password.

This is caused by an early return in model/submodules/brute_force_protection.rb:68 which skips the update of the `lock_expires_at` field. So failed attempts don't update `lock_expires_at` as long as `lock_expires_at` contains any date, even if already passed.

I'll provide a fix.

athix:
I could be mistaken, but this sounds like a misunderstanding of how the brute force protection works.

My understanding of the functionality, is that if an account receives X invalid login attempts within a certain time frame _all_ login attempts will be blocked until the login_lock_time_period expires. Once that time frame has passed, logins will be allowed as usual unless the bad login count is exceeded again. This process can repeat infinitely, and mainly acts as a way to slow down brute force attempts to the point of in-feasibility, while not overly burdening users.

Extending a lockout period if additional bad requests are received during the lockout is _not_ an intended functionality, as far as I'm aware. The hook that returns if it's present on line 68 is intentional, as the `lock_expires_at` field will be made `nil` once a lock expires, using the callback on [line 121](https://github.com/Sorcery/sorcery/blob/master/lib/sorcery/model/submodules/brute_force_protection.rb#L121).

futuretap:
You misunderstood the point. After the initial login_lock_time_period expires, an infinite number of brute force attempts are allowed without ever considering another login_lock_time_period. 

This effectively disables the brute force protection after the first login_lock_time_period has passed. It is only enabled back again after a successful login (at which point, the lock_expires_at field is set back to nil).

This is clearly a severe vulnerability.

athix:
@futuretap 

[`base.sorcery_config.before_authenticate << :prevent_locked_user_login`](https://github.com/Sorcery/sorcery/blob/master/lib/sorcery/model/submodules/brute_force_protection.rb#L36)

Inside `def prevent_locked_user_login`:

[Lines 120-123:](https://github.com/Sorcery/sorcery/blob/master/lib/sorcery/model/submodules/brute_force_protection.rb#L120)

```ruby
if !login_unlocked? && config.login_lock_time_period != 0
  login_unlock! if send(config.lock_expires_at_attribute_name) <= Time.now.in_time_zone
end
```

[...]

PR #232, opened 2020-04-30
Fix for brute_force vulnerability #231

futuretap:
extend the account lock for subsequent failed login attempts after the initial lock period

### DIFF
--- a/lib/sorcery/controller/submodules/brute_force_protection.rb
+++ b/lib/sorcery/controller/submodules/brute_force_protection.rb
@@ -26,7 +26,10 @@ module Sorcery
           # Runs as a hook after a failed login.
           def update_failed_logins_count!(credentials)
             user = user_class.sorcery_adapter.find_by_credentials(credentials)
-            user.register_failed_login! if user
+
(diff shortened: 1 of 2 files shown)


athix:
Considering this changes functionality in a potentially undesirable way, we'll want to implement this as a configuration option that defaults to `false`. See comments on #231

futuretap:
While #231 is a clear and severe vulnerability (see my new comment), we could of course discuss what the best fix is. IMO, there are two resolution options:

1. When `lock_expires_at` has passed, reset the failed login counter back to 0 and set `lock_expires_at` to nil, effectively giving the user another full set of attempts.
2. When `lock_expires_at` has passed, extend the lock immediately for another `login_lock_time_period` at the first failed login attempt (if there wasn't a successful login in between).

I've opted for option 2, the more secure option. But option 1 would be certainly fine, too, and we could change this. I recommend against a config option which is hard to explain and understand.

athix:
Option 1 is already implemented, see comments on issue.

futuretap:
OK, would be a reasonable option and explanation!

Thanks for quickly fixing the vulnerability!

PR #237, opened 2020-05-08
Updated the generators to allow name spaced models to be passed in.

crimson-knight:
Example: "ExampleNameSpace::User" parameter which will generate a table migration with "example_name_space_user" as the table name.

It's my first shot at helping out with some open source work, so I hope this is a useful starting contribution!


kentarohorie:
@athix Could you review this PR?

PR #238, opened 2020-05-10
Add support for Rails 6

sinsoku:
It adds Rails 6 to TravisCI.

This pull request also contains two changes to fix failing tests with Rails 6.

- Since the usage of `MigrationContext` was changed in rails/rails#36439, adds a branch to MigrationHelper.

- We are loading 'rails/all' in our test code, it implicitly load action_text from Rails 6. ActionText requires ApplicationController, adds it.
  - ref: https://github.com/rails/rails/blob/v6.0.3/actiontext/lib/action_text/engine.rb#L50

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: ruby
+cache: bundler
+
 rvm:
   - 2.4.9
   - 2.5.7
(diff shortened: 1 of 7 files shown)


athix:
Thank you @sinsoku!

PR #239, opened 2020-05-11
JWT authentication

alexeyramazanov:
As it's written - JWT support for sorcery.

Partially based on #70 and #167.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,15 @@ User.load_from_activation_token(token)
 @user.activate!
 ```
 
+### JWT Authentication
+
+```ruby
+require_jwt_authentication # This is a before action
+jwt_authenticate(email, password) # => {token: 'token', payload: {..}}
+jwt_from_header # Token extracted from header
+jwt_decoded_payload # Payload extracted from token
+```
+
 Please see the tutorials in the github wiki for detailed usage information.
 
 ## Installation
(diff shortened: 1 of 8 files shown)


sqlninja (commented):
As someone that is currently utilizing https://github.com/hayfever/sorcery-jwt for tokens, I would love to see this move forward. I currently have a snag with the mentioned gem and oauth. When a user logs in with a provider (i.e. Twitter), I am not getting a token. I suspect JWT being "baked" into Sorcery would allow for that...

athix:
@sqlninja I'm about to look into this for Sorcery v1, one thing that I would like to know is if the community would prefer this to be a part of `sorcery-core`, or keep this as a plugin you can add `sorcery-jwt`. I'm leaning towards the latter, seeing as most applications don't need JWT.

To @alexeyramazanov, @hayfever, @wilddima, @ebihara99999: would any of you be willing to assist me in implementing this in Sorcery v1? Being pointed at the right documentation for JWT, as well as having the perspective of someone who already uses it would both be incredibly helpful for me. Whatever we design, I want to ensure it is interoperable with other plugins (e.g. `sorcery-oauth`).

hayfever:
I would agree that a plugin seems to make the most sense - I enjoy the modularity of sorcery even if a lot of the plugins end up getting used as "defaults".

I would definitely be willing to help where needed - I will say that I haven't been using my extension for quite a while as I had written it solely to authenticate a mobile app in a previous project I was working on. The project I'm now working on (unfortunately) uses devise so I've been out of the sorcery world for a little while, too.

athix:
I created a new issue on the rework repo for discussing the JWT implementation, anyone who wants to help is welcome to comment there: Sorcery/sorcery-rework#9

Issue #240, opened 2020-05-12
Ruby 2.7.1 deprecation warnings


kapso:
Seeing these warnings after upgrading to Ruby 2.7.1

### Configuration

- Sorcery Version: `0.15.0`
- Ruby Version: `2.7.1`
- Framework: `Rails 6.0.3`

Warning:
```
/app/vendor/bundle/ruby/2.7.0/gems/sorcery-0.15.0/lib/sorcery/adapters/active_record_adapter.rb:38: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/app/vendor/bundle/ruby/2.7.0/gems/activemodel-6.0.3/lib/active_model/callbacks.rb:144: warning: The called method is defined here
```

athix:
[active_record_adapter.rb#L38](https://github.com/Sorcery/sorcery/blob/master/lib/sorcery/adapters/active_record_adapter.rb#L38)

[Potentially related blog post](https://piechowski.io/post/last-arg-keyword-deprecated-ruby-2-7/)

rails/rails#38202

athix:
Sounds like the solution is changing L38 to the following?

```ruby
@klass.send "#{time}_#{event}", method_name, **options.slice(:if, :on)
```

athix:
This should be resolved on current master.

PR #241, opened 2020-05-12
Fix keyword arguments warnings on Ruby 2.7

sinsoku:
It fixes codes to match method signatures in rails and bcrypt.

## See also

- [active_record/persistence.rb#L469](https://github.com/rails/rails/blob/v6.0.3/activerecord/lib/active_record/persistence.rb#L469)
- [active_model/callbacks.rb](https://github.com/rails/rails/blob/v6.0.3/activemodel/lib/active_model/callbacks.rb)
- [bcrypt/password.rb#L43](https://github.com/codahale/bcrypt-ruby/blob/v3.1.13/lib/bcrypt/password.rb#L43)


bspellacy:
Hey Team! Do we have an expected timeline for this to PR to get reviewed / released? We just finished our ruby 2.7 upgrade and `sorcery` is the last gem on the list 😃

Thank you!

cc @Ch4s3

athix:
Fixes #240

athix:
CI seems to be borked, but the tests passed previous to the CHANGELOG update, and that should have no effect on any code. Merging into master.

PR #250, opened 2020-06-28
Support overwriting config per controller and multi-class logins

sinsoku:
This pull request is intended to support logins for multiple roles, such as User and Admin.

First, change the configuration to use an instance instead of Module to allow overwriting it per controllers.
In a controller, you can access it with `sorcery_config`.

Next, add `login_session_key` to the configuration so that we can change the session key from `:user_id`.
It also supports Proc to be able to change the session key automatically according to the `user_class`.

These changes allow you to maintain multiple sessions.

```ruby
class Admins::BaseController < ApplicationController
  before_action :set_admin_to_sorcery_config

  private

  def set_admin_to_sorcery_config
    sorcery_config.user_class = Admin
    sorcery_config.login_session_key = :admin_id
  end
end
```

## NOTE

I'd like to get opinions from Sorcery maintainers, as being able to override config per controllers is the notable design change.
Overriding the config is a big change, so we might to separate pull requests.

### DIFF
--- a/lib/sorcery/controller/config.rb
+++ b/lib/sorcery/controller/config.rb
@@ -1,39 +1,36 @@
 module Sorcery
   module Controller
     module Config
-      class << self
-        attr_accessor :submodules
+      DEFAULTS = {
         # what class to use as the user class.
-        attr_accessor :user_class
+        :user_class                           => nil,
+        :submodules                           => [],
         # what controller action to call for non-authenticated users.
-        attr_accessor :not_authenticated_action
+        :not_authenticated_action             => :not_authenticated,
+        :login_sources                        => Set.new,
+        :after_login                          => Set.new,
+        :after_failed_login                   => Set.new,
+        :before_logout                        => Set.new,
+        :after_logout                         => Set.new,
+        :after_remember_me                    => Set.new,
         # when a non logged in user tries to enter a page that requires login,
         # save the URL he wanted to reach, and send him there after login.
-        attr_accessor :save_return_to_url
+        :save_return_to_url                   => true,
         # set domain option for cookies
-        attr_accessor :cookie_domain
+        :cookie_domain                        => nil
+      }.freeze
+      private_constant :DEFAULTS
 
-        attr_accessor :login_sources
-        attr_accessor :after_login
-        attr_accessor :after_failed_login
-        attr_accessor :before_logout
-        attr_accessor :after_logout
-        attr_accessor :after_remember_me
+      class << self
+        def add_defaults(defaults)
+          singleton_class.attr_accessor(*defaults.keys)
+          @defaults.merge!(
+            defaults.transform_keys { |key| "@#{key}".to_sym }
+          )
+        end
 
         def init!
-          @defaults = {
-            :@user_class                           => nil,
-            :@submodules                           => [],
-            :@not_authenticated_action             => :not_authenticated,
-            :@login_sources                        => Set.new,
-            :@after_login                          => Set.new,
-            :@after_failed_login                   => Set.new,
-            :@before_logout                        => Set.new,
-            :@after_logout                         => Set.new,
-            :@after_remember_me                    => Set.new,
-            :@save_return_to_url                   => true,
-            :@cookie_domain                        => nil
-          }
+          @defaults = DEFAULTS.map { |k, v| ["@#{k}".to_sym, v.dup] }.to_h
         end
 
         # Resets all configuration options to their default values.
@@ -63,6 +60,7 @@ module Sorcery
       end
 
       init!
+      add_defaults(DEFAULTS)
       reset!
     end
   end
--- a/lib/sorcery/controller/config.rb
+++ b/lib/sorcery/controller/config.rb
@@ -17,7 +17,8 @@ module Sorcery
         # save the URL he wanted to reach, and send him there after login.
         :save_return_to_url                   => true,
         # set domain option for cookies
-        :cookie_domain                        => nil
+        :cookie_domain                        => nil,
+        :login_session_key                    => :user_id
       }.freeze
       private_constant :DEFAULTS
 
@@ -90,6 +91,14 @@ module Sorcery
 
       def dup
         self.class.new(attributes)
+      end
+
+      def session_key
+        if login_session_key.is_a?(Proc)
+          login_session_key.call(self)
+        else
+          login_session_key
+        end
       end
 
       private def attributes
--- a/lib/sorcery/controller/config.rb
+++ b/lib/sorcery/controller/config.rb
@@ -1,6 +1,6 @@
 module Sorcery
   module Controller
-    module Config
+    class Config
       DEFAULTS = {
         # what class to use as the user class.
         :user_class                           => nil,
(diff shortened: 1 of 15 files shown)


athix:
@sinsoku Honestly, I'm looking at starting a rewrite of Sorcery from scratch, so maybe it would make more sense for us to discuss how to implement these changes with the new plugin and config system instead. @Fustrate made a gist with one potential way we could rework the config to be clearer/less bloated, so I would take a look at that and see how it would fit in with the changes you're wanting to make.

crimson-knight:
@athix an entire rewrite from scratch? I saw the repo and followed it. That sounds incredibly lofty, and like it would be dangerous to transition to. I have all of my projects using Sorcery and I'd rather not have to do some kind of big rework to keep using it if 1.x comes out with breaking changes.

What's the reasoning behind a full rewrite vs just refactoring in stages?

crimson-knight:
@sinsoku I was thinking about solving this problem using a simple permissions-based system. I don't have a rock-solid solution yet but I'm constantly bumping into it with my current projects to the point that I will need to take a few days and make something solid that I can use now and preferably re-use in the future.

athix:
@crimson-knight I would say that it sounds a lot scarier on paper than what my intentions with it actually are. Essentially the rewrite is more to clear out 10 years of work-arounds and start with a fresh base with good testing from square 0, than anything else. The migration plan for existing apps would be:

* Change gem requirements
  * `sorcery` -> `sorcery-core`
  * Add `sorcery-oauth` if you're using the `External` submodule
* Update your config (the wiki will include a guided how-to with examples)
* Find and replace method names (this may not be needed, depending on if method names make sense as-is)
* Done


Getting a Sorcery v1 out that reworks some of the major pain points we've been seeing has been on the docket for years now. Just finally making the commitment to do it now.

athix:
@sinsoku @crimson-knight for permissions, have you looked into using the `pundit` gem? It works well with Sorcery, and may provide some of the functionality you're looking for.

wout:
Isn't the intention of this PR to add multi-model authentications to allow two independent log-in flows eventually? I've been looking into that as well.

Currently, Sorcery allows defining only one model (e.g. `User`). But what if your app needs two (or more) completely isolated login systems?

For example, a shopping cart where shop owners can log in to their admin panel, and on the other side, a frontend where their customers can log in to manage their account with that particular shop. Throwing everything in one table isn't really ideal in such a scenario.

Maybe something to consider for the rewrite? :)

athix:
@wout Ahhh, okay that makes a lot more sense than what I was assuming, interesting. Yeah that sounds worth considering for sure, thanks for clarifying that!

So this is less the authorization aspect, and more supporting multiple separate login tables?

crimson-knight:
@wout that's easy to do though, and doesn't necessarily require an entirely new or different login system. I've already done this without a ton of extra setup/work and it's been pretty flexible. I guess I can add that to my documentation list 😅 🤷‍♂️

wout:
> So this is less the authorization aspect, and more supporting multiple separate login tables?

@athix Not sure, but from what's in the description I suppose that's the case with this PR.

> I guess I can add that to my documentation list.

@crimson-knight That would be great! 😀️ I'll need something like that eventually for two apps I'm working on, so I'd love to see how it's done.

lib/sorcery/controller/config.rb
```diff
@@ -1,69 +1,112 @@
...
-          @defaults.each do |k, v|
-            instance_variable_set(k, v)
-          end
-        end
+        def add_defaults(defaults)
+          attr_accessor(*defaults.keys)
+
+          # Delegate accessor methods to instance
+          defaults.each_key do |name|
+            class_eval <<-RUBY, __FILE__, __LINE__ + 1
```
athix, line 39: @sinsoku How does using a multiline string with `, __FILE__, __LINE__ + 1` differ from using a do/end block for class_eval? I'm assuming there's a reason behind doing it this way, but I'm not familiar enough to figure out what it is.

> Evaluates the string or block in the context of mod, except that when a block is given, constant/class variable lookup is not affected

Guessing this might be related, but still haven't grokked it yet.

athix:
@sinsoku I'm now at the configuration part of the refactor, are you available to discuss this sometime?

I'm considering making the configuration a centralized class rather than splitting it between controller and model, and updating it so that you can do custom configuration as a part of the `authenticates_with_sorcery!` call. Running into some issues with how the model config differs from the controller, and am considering if that's because of a bad approach, or just because previously only the model config was configurable per class.

Essentially how I'm thinking it would work:

* Defaults loaded in from the sorcery code.
* Initializer overwrites defaults with user's preferred settings. In most apps, this will be the only config needed.
* `authenticates_with_sorcery! do |config|` will yield with the config instance for the calling class, either the model or controller depending on where it's called from. This is how you would change settings on a per class basis. Using this block is optional, and will replace any settings from the initializer if they conflict.
* Once the yield is over, `authenticates_with_sorcery!` will continue with the rest of the usual include/extend process, using the config instance specific to that class.

I'll try out some ideas based off of your PR here. Please let me know if you would be interested in helping.

athix:
Theoretically this should now be implemented on the v1 rework.

lib/sorcery/controller/config.rb
sinsoku: @athix I wrote it according to the following code.
https://github.com/Sorcery/sorcery/blob/a02c1247642357129ea739354c22978a06fccaa9/lib/sorcery/controller/submodules/external.rb#L40-L44

I was interested in the differences, so I investigated and found that string performed better than block.

```ruby
require 'benchmark/ips'

class A
  class << self
    def instance
      "dummy"
    end

    class_eval do
      define_method("with_block") { instance }
    end

    class_eval <<-RUBY, __FILE__, __LINE__ + 1
      def #{"with_str"}
        instance
      end
    RUBY
  end
end

Benchmark.ips do |x|
  x.report("with_block") { A.with_block }
  x.report("with_str") { A.with_str }

  x.compare!
end
```

```
Warming up --------------------------------------
          with_block   714.383k i/100ms
            with_str     1.005M i/100ms
Calculating -------------------------------------
          with_block      7.705M (± 2.8%) i/s -     38.577M in   5.011016s
            with_str     10.466M (± 2.1%) i/s -     53.241M in   5.089350s

Comparison:
            with_str: 10466030.0 i/s
          with_block:  7704795.0 i/s - 1.36x  (± 0.00) slower
```

sinsoku:
I'm sorry, the reply was delayed.

> Isn't the intention of this PR to add multi-model authentications to allow two independent log-in flows eventually? I've been looking into that as well.

That's right.

My use case required two log-in flows for users and administrators.
I've tried overriding the config per `/users/login` and `/admins/login` paths, but I couldn't because `Sorcery::Controller::Config` is not thread safe.

This pull request is intended to make config thread-safe so that we can override the config per request.
I want to change not only the model but also other settings such as the session timeout and the redirect path when login fails depending on the login flow.

lib/sorcery/controller/config.rb
athix: Ahh, so it's a performance improvement. Interesting :slightly_smiling_face:

Issue #227, opened 2020-04-13
Master version fails with "To use reset_password submodule, you must define a mailer"


Hirurg103:
### Configuration

- Sorcery Version: [c30cefa7](https://github.com/Sorcery/sorcery/commit/c30cefa751c364ee20b361eb63a766f782b69e5e)
- Ruby Version: `2.5.1`
- Framework: `Rails`, versions `6.0.2.2`, `5.2.4`
- Platform: `Mac OS Mojave 10.14.6`

### Expected Behavior

No error `To use reset_password submodule, you must define a mailer` in the rails development console

### Actual Behavior

Any operation in the rails development console fails with 

```
validate_mailer_defined': To use reset_password submodule, you must define a mailer (config.reset_password_mailer = YourMailerClass). (ArgumentError)
```

### Steps to Reproduce

There is code snippet that reproduces the issue:

```ruby
# issue_with_sorcery.rb
require "bundler/inline"

gemfile(false) do
  gem 'rails', '6.0.2.2'
  gem 'sorcery', github: 'Sorcery/sorcery', branch: 'master', require: false
  gem 'sqlite3', '1.4.2'
end

require 'rails/all'
require 'sorcery'
module TestApp
  class Application < Rails::Application
    config.load_defaults 6.0
  end
end

ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
ActiveRecord::Base.logger = Logger.new(STDOUT)
ActiveRecord::Schema.define do
  create_table "users", force: :cascade do |t|
    t.string "email", null: false
    t.string "name"
  end
end

Rails.application.config.sorcery.submodules = [:reset_password]

class UserMailer < ActionMailer::Base
  def reset_password_email(user)
    mail(to: user.email, subject: 'Welcome') do |f|
      f.html { render plain: "Hi Dear #{user.name}" }
    end
  end
end

Rails.application.config.sorcery.configure do |config|
  config.user_config do |user|
    user.reset_password_mailer = UserMailer
    config.user_class = "User"
  end
end

class User < ActiveRecord::Base
  authenticates_with_sorcery!
end

u = User.create(email: 'jane@example.org', name: 'Jane')
puts u.attributes
```
It is expected that this script will print user attribute values when you run it with `ruby issue_with_sorcery.rb` but it fails with
```
[...]

Issue #132, opened 2018-06-20
Loads ActionController::Base in Rails applications always


andrewhood125:
Rails has an `--api` option that changes the default parent of the ApplicationController from ActionController::Base to ActionController::API. 

That class has some good documentation on why you would not want to load the entire ActionController::Base class.

https://github.com/rails/rails/blob/94360451eae2701fc04d6b132c44d9622bbe0cb5/actionpack/lib/action_controller/api.rb#L8-L11

In the engine: https://github.com/Sorcery/sorcery/blob/master/lib/sorcery/engine.rb#L11-L13

ActionController::Base is explicitly referenced. In a rails app that has been configured this is the first time that file will have been referenced and it will now be loaded. During load it also initializes a number of modules including RequestForgeryProtection.

athix:
I believe this has been fixed by the on_load calls.

PR #255, opened 2020-08-10
Add back changes that fix deprecation warning related to autoloading in Rails 6

Hirurg103:
This PR introduces the following changes:

1. Add back changes from #209 that fix deprecation warning "Initialization autoloaded the constants ActionText::ContentHelper and ActionText::TagHelper."

2. Call sorcery config block where it is defined to fix "To use reset_password submodule, you must define a mailer" error. This error started to occur after adding changes from #209 when calling the user model in the rails development console.

### DIFF
--- a/lib/sorcery/controller/config.rb
+++ b/lib/sorcery/controller/config.rb
@@ -54,11 +54,8 @@ module Sorcery
         end
 
         def configure(&blk)
-          @configure_blk = blk
-        end
-
-        def configure!
-          @configure_blk.call(self) if @configure_blk
+          update!
+          blk.call(self) if blk
         end
       end
 
--- a/lib/sorcery/engine.rb
+++ b/lib/sorcery/engine.rb
@@ -9,18 +9,12 @@ module Sorcery
 
     # TODO: Should this include a modified version of the helper methods?
     initializer 'extend Controller with sorcery' do
-      # FIXME: on_load is needed to fix Rails 6 deprecations, but it breaks
-      #        applications due to undefined method errors.
-      # ActiveSupport.on_load(:action_controller_api) do
-      if defined?(ActionController::API)
-        ActionController::API.send(:include, Sorcery::Controller)
+      ActiveSupport.on_load(:action_controller_api) do
+        ActionController::API.include(Sorcery::Controller)
       end
 
-      # FIXME: on_load is needed to fix Rails 6 deprecations, but it breaks
-      #        applications due to undefined method errors.
-      # ActiveSupport.on_load(:action_controller_base) do
-      if defined?(ActionController::Base)
-        ActionController::Base.send(:include, Sorcery::Controller)
+      ActiveSupport.on_load(:action_controller_base) do
+        ActionController::Base.include(Sorcery::Controller)
         ActionController::Base.helper_method :current_user
         ActionController::Base.helper_method :logged_in?
       end
--- a/lib/sorcery/controller.rb
+++ b/lib/sorcery/controller.rb
@@ -14,8 +14,6 @@ module Sorcery
           # rubocop:enable Lint/HandleExceptions
         end
       end
-      Config.update!
-      Config.configure!
     end
 
     module InstanceMethods
--- a/spec/controllers/controller_activity_logging_spec.rb
+++ b/spec/controllers/controller_activity_logging_spec.rb
@@ -45,6 +45,8 @@ describe SorceryController, type: :controller do
       login_user(user)
       now = Time.now.in_time_zone
       Timecop.freeze(now)
+
+      sorcery_controller_property_set(:register_logout_time, true)
       expect(user).to receive(:set_last_logout_at).with(be_within(0.1).of(now))
 
       logout_user


lib/sorcery/engine.rb
```diff
@@ -9,18 +9,12 @@ class Engine < Rails::Engine
...
+        ActionController::API.include(Sorcery::Controller)
       end
 
-      # FIXME: on_load is needed to fix Rails 6 deprecations, but it breaks
-      #        applications due to undefined method errors.
-      # ActiveSupport.on_load(:action_controller_base) do
-      if defined?(ActionController::Base)
-        ActionController::Base.send(:include, Sorcery::Controller)
+      ActiveSupport.on_load(:action_controller_base) do
+        ActionController::Base.include(Sorcery::Controller)
```
athix, line 17: This solution is confirmed as breaking for existing applications, although an alternative that does work is on the v1 rework.

lib/sorcery/controller/config.rb
```diff
@@ -54,11 +54,8 @@ def user_config(&blk)
...
 
         def configure(&blk)
-          @configure_blk = blk
-        end
-
-        def configure!
-          @configure_blk.call(self) if @configure_blk
+          update!
+          blk.call(self) if blk
         end
```
athix, line 59: With how the config works in v1, I don't know that immediately calling the block makes sense. `configure` essentially sets the global config, which gets executed on duplicates rather than the primary instance. I'll definitely keep this in mind though, and see if it can be simplified to a single method when testing.

athix:
@Hirurg103 you might want to take a look at Sorcery/sorcery-rework. The bulk of the rework is changing how Sorcery handles the config and plugin (submodule) loading, which fixes the deprecation warnings as a consequence.

athix:
Also important to note, the `defined?` calls have been removed everywhere that I've found them. It causes some pretty nasty load order bugs, potentially related to the issues you've had in the past.

lib/sorcery/engine.rb
Hirurg103: Do you mean #227 there? Or does this change cause other issues?

athix: Per FIXME, changing to on_load causes existing applications to generate undefined method errors. This has been pretty consistent every time we've tried switching.

lib/sorcery/controller/config.rb
Hirurg103: > `configure` essentially sets the global config, which gets executed on duplicates rather than the primary instance.

While inspecting the code I found out that `Sorcery::Controller::Config` class (not instance) is intended to be used in controllers and it is not relevant where we set class variables (in the `configure!` method after action controller is laded or in the sorcery initializer when calling `configure` with a block)

Instances of `Sorcery::Model::Config` [are configured in `::Sorcery::Controller::Config.user_config` block](https://github.com/Sorcery/sorcery/blob/v0.14.0/lib/sorcery/model.rb#L20) and used in user models

There is another way to fix #227: move configure user block out of the configure controller block. Sorcery initializer will look like this:

```ruby
# Rails.application.config.sorcery.configure is a way too long, 
# Sorcery.configure looks more elegant
Sorcery.configure_controller do |config|
  ...
end

Sorcery.configure_user do |config|
  ...
end

```

Hirurg103:
Thank you for pointing me to Sorcery/sorcery-rework and fixing deprecation warnings there. Shall I close this PR?

athix:
@Hirurg103 we can leave it open for now, it serves as a good reminder for me to test for method undefined errors, as well as deprecation warnings. I plan on closing out all the old PRs that are marked `implemented in v1` once v1 gets released.

lib/sorcery/controller/config.rb
athix: I definitely like calling `Sorcery.configure` over `Rails.application.config.sorcery.configure`, I'll see about implementing that for sure.

I've reworked how the config works quite a bit in the rework, but if there ends up still being a `configure_user` block, I'll extract it out like you've suggested. I like that separation better.

lib/sorcery/engine.rb
Hirurg103: Are there any other [Sorcery/sorcery issues](https://github.com/Sorcery/sorcery/issues) related to switching to the `on_load` syntax other than #227? If yes, could you please point me to them?

Hirurg103: @athix [you say](https://github.com/Sorcery/sorcery/pull/255#issuecomment-671629664) that the deprecation warning from autoloading in Rails 6 (#209) has been fixed in Sorcery/sorcery-rework. When it is not going to be fixed in the main gem, then the question above is irrelevant and you can ignore it

athix: #132 is related

athix: > When it is not going to be fixed in the main gem, then the question above is irrelevant and you can ignore it

The rework (`sorcery-core`) will become the main gem once v1 ships, so in that sense it should be resolved in the main gem. As for back-porting it to the current `0.15.X` code-base, my free time is already completely consumed by the v1 rework so it's unlikely to happen until after `1.0.0` is live regardless.

Honestly, it would probably make sense for you to create a temporary fork with your changes here and use that until v1 is released. `0.15` shouldn't be receiving any updates in the meantime, so you won't have to worry about pulling down upstream changes until v1 goes live and you can cut over to that version.

Hirurg103: Ok, got it. Thank you for the advice @athix

PR #260, opened 2020-10-26
Add Battlenet Provider

jojoa:
I Added BattleNet as OAuth Provider as documented in theri api docu and tested it local.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Changelog
 ## HEAD
 
-* Add BattleNet Provider [#258](https://github.com/Sorcery/sorcery/pull/258)
+* Add BattleNet Provider [#260](https://github.com/Sorcery/sorcery/pull/260)
 * Fix ruby 2.7 deprecation warnings [#241](https://github.com/Sorcery/sorcery/pull/241)
 
 ## 0.15.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 ## HEAD
 
+* Add BattleNet Provider [#258](https://github.com/Sorcery/sorcery/pull/258)
 * Fix ruby 2.7 deprecation warnings [#241](https://github.com/Sorcery/sorcery/pull/241)
 
 ## 0.15.0
--- a/lib/generators/sorcery/templates/initializer.rb
+++ b/lib/generators/sorcery/templates/initializer.rb
@@ -227,15 +227,16 @@ Rails.application.config.sorcery.configure do |config|
   # config.line.user_info_mapping = {name: 'displayName'}
 
   
-  # For infromation about Discord API
+  # For information about Discord API
   # https://discordapp.com/developers/docs/topics/oauth2
   # config.discord.key = "xxxxxx"
   # config.discord.secret = "xxxxxx"
   # config.discord.callback_url = "http://localhost:3000/oauth/callback?provider=discord"
   # config.discord.scope = "email guilds"
 
-  # For infromation about BattleNet API
+  # For inforation about Battlenet API
   # https://develop.battle.net/documentation/guides/using-oauth
+  # config.battlenet.site = "https://eu.battle.net/" #See Website for other Regional Domains
   # config.battlenet.key = "xxxxxx"
   # config.battlenet.secret = "xxxxxx"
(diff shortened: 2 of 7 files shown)


athix (commented):
Looks good at first glance! I'll take a closer look at this when I can and merge it into master.

lib/sorcery/providers/battlenet.rb
```diff
@@ -0,0 +1,53 @@
...
+    class Battlenet < Base
+      include Protocols::Oauth2
+
+      attr_accessor :auth_path, :scope, :token_url, :user_info_path
+
+      def initialize
+        super
+
+        @scope          = 'openid'
+        @site           = 'https://eu.battle.net/'
```
athix, line 14: Will defaulting to the EU subdomain of battlenet cause issues for users in the US or other server areas?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 ## HEAD
 
+* Add BattleNet Provider [#258](https://github.com/Sorcery/sorcery/pull/258)
```
athix, line 4: The PR number here should be updated to #260.

lib/generators/sorcery/templates/initializer.rb
```diff
@@ -226,12 +226,20 @@
...
 
+  
   # For infromation about Discord API
   # https://discordapp.com/developers/docs/topics/oauth2
   # config.discord.key = "xxxxxx"
   # config.discord.secret = "xxxxxx"
   # config.discord.callback_url = "http://localhost:3000/oauth/callback?provider=discord"
   # config.discord.scope = "email guilds"
+
+  # For infromation about BattleNet API
```
athix, line 237: Minor typo, `information`

Also it might be good to change `BattleNet` to `Battlenet` so we're consistent with the casing.

lib/sorcery/providers/battlenet.rb
```diff
@@ -0,0 +1,53 @@
...
+      # to get authenticated at the external provider's site.
+      def login_url(_params, _session)
+        authorize_url(authorize_url: auth_path)
+      end
+
+      # tries to login the user from access token
+      def process_callback(params, _session)
+        args = {}.tap do |a|
+          a[:code] = params[:code] if params[:code]
+        end
```
athix, line 41: Does passing `code: nil` as an argument cause issues for the battlenet endpoint? This could potentially be simplified as `args = { code: params[:code] }` if that's not a concern.

athix (changes requested):
Some minor things I noticed.

lib/sorcery/providers/battlenet.rb
jojoa: i actually forgot this, i planned to implement a option for this. but im not sure if it is critical maybee only some higher ping. i will implement the option anyway sor setting the location

lib/sorcery/providers/battlenet.rb
jojoa: what do you mean exactly? If you dont pass the code to the api you get an http 400.

athix (changes requested):
@jojoa looks good other than the `eu.battle.net` concern. Do you have time to look into their documentation? A link to their docs where it talks about which subdomains to use would be helpful for me to better understand their implementation of OAuth2.

jojoa:
> @jojoa looks good other than the `eu.battle.net` concern. Do you have time to look into their documentation? A link to their docs where it talks about which subdomains to use would be helpful for me to better understand their implementation of OAuth2.

I ve tested it. i can succesfully use the us domain with my eu account. so i think there are no problems. the doc says that only CN has an seperate Data Department. But i am not sure how this could be implemented. also the game infos could be different, but the general login info is same. so i dont think it is important for an auth Gem to have things like the Wow chars.

jojoa:
https://develop.battle.net/documentation/guides/regionality-and-apis

jojoa:
any news here?

athix (approved):
If anything is broken, it should only affect people using this specific provider and can be fixed with a later update. Going to go ahead and merge this for now.

athix:
@jojoa You should be able to use this by pointing your Gemfile at the master branch of Sorcery via Github. I'm planning on focusing on the v1 release of Sorcery over the next week, so I probably won't cut a new release until after that's ready to deploy.

PR #261, opened 2020-11-19
Change/variable name

kentarohorie:
I want to add changing about two point. It's simple refectoring.

- Change variable name in add_provider_to_user because it was wrong.
- Make add_provider_to_user's return simple because it was unneeded.

### DIFF
--- a/lib/sorcery/model/submodules/external.rb
+++ b/lib/sorcery/model/submodules/external.rb
@@ -100,10 +100,8 @@ module Sorcery
                                                  sorcery_config.provider_attribute_name => provider)
               authentication.sorcery_adapter.save(validate: false)
             else
-              authentication = false
+              false
             end
-
-            authentication
           end
         end
       end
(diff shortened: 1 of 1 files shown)


athix (approved):
Needs some housekeeping stuff, but the code change itself looks good.

lib/sorcery/model/submodules/external.rb
```diff
@@ -96,14 +96,12 @@ def add_provider_to_user(provider, uid)
...
             # first check to see if user has a particular authentication already
             if sorcery_adapter.find_authentication_by_oauth_credentials(authentications, provider, uid).nil?
-              user = send(authentications).build(sorcery_config.provider_uid_attribute_name => uid,
+              authentication = send(authentications).build(sorcery_config.provider_uid_attribute_name => uid,
                                                  sorcery_config.provider_attribute_name => provider)
-              user.sorcery_adapter.save(validate: false)
+              authentication.sorcery_adapter.save(validate: false)
             else
-              user = false
+              false
```
athix, line 103: Actually, looking closer at this, I think this changes the return value when the authentication is present. Previously it would return `authentication`, but with this change it would instead return `authentication.sorcery_adapter.save(validate: false)` as that's the last value called in the method.

Can you revert the second part of this, and keep it to just renaming the variable to be more accurate?

athix (changes requested):

PR #267, opened 2021-01-02
Test against ruby 3

driv3r:
## Changes:
- test against Ruby 3.0
- extend GH Workflow file to include job matrix
- cache gems in workflows

## Follow up:
- remove `.travis.yml` file

## Notes:
- there's no functionality change, so no Changelog entry

- [x] Description of changes


athix:
With the addition of Github actions, I'm heavily leaning towards removing TravisCI as it's had quite a few issues over the years that I've had to debug. Adding tests for Ruby 3 makes sense though, so I'll look into testing it with the Github action (if it isn't already doing so).

driv3r:
@athix if you want i could update PR to have github workflow 😉

athix:
@driv3r That would be great :smile:

driv3r:
@athix from what I can see, there's already a workflow file, but it only runs against `master` branch, should I also update it to run against PRs?

athix:
@driv3r as far as I can tell, it does run against PRs, using the master branch as its comparison? See config:

```yaml
on:
  push:
    branches: [ master ]
  pull_request:
    branches: [ master ]
```
That said, I haven't played around with the Github actions config too much. One thing I'm not sure about is if it would be better/possible to run ruby 2 and ruby 3 on the same action, or if they should be separated out into two actions.

If you would like to explore that, I would appreciate the help! I'm currently focused on the v1 rework as far as my time towards Sorcery goes.

driv3r:
I'll try to tackle this today

driv3r:
@athix we will have to wait with the results due to 
image

When it's running and this PR is merged, I would suggest removing travis in followup PR after switching the required checks to github actions checks. I'll add a `meta` step that will require all tests from matrix, so you could have a single `required` check in settings (instead of selecting every single run from matrix)

driv3r:
@athix works, also shows that ruby 3.0 won't work together with rails 5.2.0, should we update patch version to 5.2.4.5 and try again?

athix:
@driv3r the fuzzy lock should use the latest patch version though: `~> 5.2.0`

Worth double checking, but it might just be that something got straight up deprecated, considering Ruby 3 came out well after Rails 6 being the latest version for a while.

athix:
Yup, checked the failing build and it is using the latest patch: `Using rails 5.2.4.5`

.github/workflows/ruby.yml
```diff
@@ -7,17 +7,46 @@ on:
...
+    env:
+      BUNDLE_GEMFILE: gemfiles/rails_${{ matrix.rails }}.gemfile
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler-cache: true
```
athix, line 43: I see we're using bundler-cache instead of the bundle install step, what changes does this do to the flow/how does it handle the caching?

driv3r: This runs `bundle install` in the scope of specified Gemfile, and caches the result with `actions/cache`

driv3r: This basically handles 3 steps:
- setup ruby
- restore/persist cache (considering Gemfile lock, ruby and platform version asking the way)
- run bundle install

athix: Thanks!

athix (approved):

PR #285, opened 2021-07-02
Update phrasing in Philosophy statement

dguardado:
Change the philosophy statement so that its heading more clearly relates to the description.

For context, it wasn't clear to me what exactly the "Magic vs Voodoo" comment was trying to say, but it definitely did not communicate to me that the lib would be easy to work on. In the worst reading, I think the comment is unnecessarily disparaging of a real religion and spiritual practice of African origin. The change in this PR is a suggestion that I think communicates the intent more clearly based on the contents of the rest of the line, but if a different change resonates better with your intent y'all are probably the experts on that.

Thanks for your consideration!

- [x] Description of changes
- [X] Changes have related RSpec tests that ensure functionality does not break

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Sorcery is a stripped-down, bare-bones authentication library, with which you ca
 
 - Less is more - less than 20 public methods to remember for the entire feature-set make the lib easy to 'get'.
 - No built-in or generated code - use the library's methods inside *your own* MVC structures, and don't fight to fix someone else's.
-- Magic yes, Voodoo no - the lib should be easy to hack for most developers.
+- Approachable and Accessible - the lib should be easy to hack for most developers.
 - Configuration over Confusion - Centralized (1 file), Simple & short configuration as possible, not drowning in syntactic sugar.
 - Keep MVC cleanly separated - DB is for models, sessions are for controllers. Models stay unaware of sessions.
 


dguardado:
I'd also like to apply this change to the wiki once approved.

athix:
I definitely agree that "Magic yes, voodoo no" is pretty unclear on what it actually means, and can be changed to something a little more clear.

I don't know that "Approachable and Accessible" fully fits the ease of modification philosophy, but I do think that would be an appropriate philosophy to enumerate, be that in a new bullet point, or by wordsmithing something that covers both the library being approachable for newcomers and easy to modify to fit a developers needs.

Perhaps we could change it to "Magically easy to modify", and add a new point for the approachability aspect for newcomers? Would love to hear ideas.

dguardado:
> Perhaps we could change it to "Magically easy to modify", and add a new point for the approachability aspect for newcomers?

This sounds good to me! I didn't necessarily mean to add anything new to the Philosophy but I also think "approachable to newcomers" is a great thing to value and would be a positive addition if that resonates with you :)

I'm happy to make those changes, or let me know how best to proceed.

athix:
What would be most helpful from my perspective is if you could brainstorm some various options and post them on this PR as a new comment.

I could then put some thought into what fits best with the library philosophies, and how Sorcery fits into the greater Ruby ecosystem, then consolidate it into an updated "Philosophy" section.

Thanks! :slightly_smiling_face:

dguardado:
Sounds good! Here are a few options I came up with to pair with the text "the lib should be easy to hack for most developers":

- "Easy to modify"
- "Adaptable"
- "Adjustable"
- "Developer friendly"
- "Powerful but familiar" (maybe getting at the feeling of the lib doing a lot for you while still being understandable)
- "Magic yes, Arcane no" (same as above, but keeps the magic wordplay)

I think the bottom half of the list could also be a good fit to cover both "ease of modification" and "approachable to newcomers". Let me know what you think!

PR #288, opened 2021-09-17
Avoid Slack users.identity lookup if access_token contains user info

jcraigk:
While testing against various Slack workspaces, I noticed that:
(1) The email and UID of the user is provided in the initial OAuth response so there is no need to call `slack.com/api/users.identity` here https://github.com/Sorcery/sorcery/blob/master/lib/sorcery/providers/slack.rb#L21. The needed info is contained in the `access_token` hash, which should probably be renamed to `access_hash` or `access_data` as it is more than a token, at least when authenticating with Slack.
(2) In some instances, Slack will respond with `{ ok: false, error: 'invalid_auth' }` when requesting user info via `slack.com/api/users.identity`. According to https://api.slack.com/methods/users.identity, this could be due to a generic auth error or restricted IP situation. In either case, the resulting error is not handled gracefully by Sorcery:

```
NoMethodError: undefined method `[]' for nil:NilClass
/gems/sorcery-0.16.1/lib/sorcery/providers/slack.rb:24:in `block in get_user_hash'
```
There is no check in the Slack Provider to see if `ok` is `true` (against the recommendation of Slack docs), and tries to access the JSON returned (`h[:user_info]['email'] = h[:user_info]['user']['email']`), which raises the `NoMethodError` above.

This PR addresses (1) directly and (2) indirectly, since the only instances I found where (2) occurred were instances where (1) was true. The proper solution may be removing the check for `slack.com/api/users.identity` altogether as it may be fully redundant. It would probably also be worth adding a check for `ok == true` and handling the error case more gracefully if the identity call fails. Without more info about why Slack is refusing to fill the request in some cases, it's probably best to leave a conditional in and try to fetch the user info if not available in the initial response.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Inline core migration index definition [#281](https://github.com/Sorcery/sorcery/pull/281)
 * Fix MongoID adapter breaking on save [#284](https://github.com/Sorcery/sorcery/pull/284)
 * Don't pass token to Slack in query params. Prevents 'invalid_auth' error [#287](https://github.com/Sorcery/sorcery/pull/287)
+* Avoid Slack users.identity lookup if access_token contains user info [#288](https://github.com/Sorcery/sorcery/pull/288)
 
 ## 0.16.1
 
--- a/lib/sorcery/providers/slack.rb
+++ b/lib/sorcery/providers/slack.rb
@@ -18,11 +18,9 @@ module Sorcery
       end
 
       def get_user_hash(access_token)
-        response = access_token.get(user_info_path)
         auth_hash(access_token).tap do |h|
-          h[:user_info] = JSON.parse(response.body)
-          h[:user_info]['email'] = h[:user_info]['user']['email']
-          h[:uid] = h[:user_info]['user']['id']
+          h[:user_info] = fetch_user_info(access_token)['user']
+          h[:uid] = h[:user_info]['id']
         end
       end
 
@@ -39,6 +37,20 @@ module Sorcery
         end
 
         get_access_token(args, token_url: token_url, token_method: :post)
+      end
+
+      private
+
+      def fetch_user_info(access_token)
+        return access_token if user_info_present?(access_token)
+
+        JSON.parse(access_token.get(user_info_path).body)
+      end
+
+      def user_info_present?(access_token)
+        access_token['user'].present? &&
+          access_token['user']['id'].present? &&
+          access_token['user']['email'].present?
       end
     end
   end


athix:
Hi @jcraigk, the conditional checking and deviation from other provider flows makes me a bit nervous, so I'd like to look into this a bit more in-depth before merging. One of the issues I'd be worried about is the conditional checks working well for your use-case, but breaking other applications that use Slack authentication slightly differently (e.g. they are pulling different data for users).

jcraigk:
@athix I share your concern. I'll inquire with Slack support to see if I can get more info on why `invalid_auth` is being returned on the `users.identity` check in this context.

Question: other than oauth user ID and email address, what other attributes does Sorcery need?

athix:
That's a good question. Typically Sorcery works within what the application developer specifies. The provider's auth UID is used for the Authentication records, and email for the User records. Beyond that, I don't know that anything is strictly required on Sorcery's end. I would look into the [create_from_provider](https://github.com/Sorcery/sorcery/blob/master/lib/sorcery/model/submodules/external.rb#L57) method and [config initializer](https://github.com/Sorcery/sorcery/blob/master/lib/generators/sorcery/templates/initializer.rb#L187) to get a better idea.

jcraigk:
Taking a look at the code, it seems like UID/email is all that is required by default. Here the user can specify a `config.slack.user_info_mapping` hash for fields beyond `email` https://github.com/Sorcery/sorcery/blob/master/lib/sorcery/model/submodules/external.rb#L57. Perhaps the solution is to skip the `user.identity` check if `config.slack.user_info_mapping` does not contain fields other than `email`? My take is the request is redundant in such a case. Also I'm thinking we should add a more graceful check for `user.identity` failure instead of letting the `NoMethodError` get raised.

Thoughts?

athix:
Could also take a look at the Omniauth strategy for Slack for inspiration (Sorcery will be moving to Omniauth in Sorcery v1, if everything goes according to plan), although it does seem to not have been updated in quite a while: https://github.com/kmrshntr/omniauth-slack/blob/master/lib/omniauth/strategies/slack.rb

jcraigk:
According to https://github.com/kmrshntr/omniauth-slack/blob/master/lib/omniauth/strategies/slack.rb#L23, it requires the team ID which means the `user.identity` check is required for that as well.

Interestingly, the original issue that spawned this thread has been resolved on Slack's side. I put in a request about the issue, they responded almost immediately asking me for details and as I was preparing that for them, I found that I was no longer receiving `invalid_auth` responses. So 🤷 maybe the issue is resolved, maybe it will come back.

athix:
@jcraigk I would say let's hold off on this PR then until you can confirm if the issue has been resolved. Saves both of us some time, and when I have time to get to working on v1, we'll likely be blowing away all this custom OAuth code anyway.

PR #290, opened 2021-10-12
Allow redirect to other host for Rails 7.0.0alpha2

rzepetor:
I give a try Rails 7.0.0alpha2 with sorcery and when trying to redirect on production server to external provider then error occurs that this redirect is unsafe and I should use ```allow_other_host: true``` option.

### DIFF
--- a/lib/sorcery/controller/submodules/external.rb
+++ b/lib/sorcery/controller/submodules/external.rb
@@ -78,7 +78,7 @@ module Sorcery
           end
 
           # get the user hash from a provider using information from the params and session.
-          def sorcery_fetch_user_hash(provider_name, access_token)
+          def sorcery_fetch_user_hash(provider_name)
             # the application should never ask for user hashes from two different providers
             # on the same request.  But if they do, we should be ready: on the second request,
             # clear out the instance variables if the provider is different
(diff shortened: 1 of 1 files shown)


lib/sorcery/controller/submodules/session_timeout.rb
```diff
@@ -36,7 +36,7 @@ def invalidate_active_sessions!
             return unless current_user.present?
 
             current_user.send(:invalidate_sessions_before=, Time.now.in_time_zone)
-            current_user.save
+            current_user.save(validate: false)
```
mladenilic, line 39: Do we need to skip validation here?

wojciech-rzepecki: If we don't, password is required. Additionally it should be independent from other validations I think

athix: I'm pretty sure I noticed this myself while working on the v1 rework. It doesn't make sense for us to care about any validations, as we're editing a value outside of those normal loops. I'll have to sanity check that before merging that in though.

Also, @wojciech-rzepecki, normally you include a proc that checks for if the password is being set or changed so that it doesn't interfere during normal editing. (e.g., a user editing their timezone)

For example:
```ruby
validates :password,
  presence: true,
  confirmation: true,
  if: -> { new_record? || changes[:crypted_password] }

validates :password_confirmation,
  presence: true,
  if: -> { new_record? || changes[:crypted_password] }
```

sebastiandeutsch:
We're running into that problem. Would it be possible to roll it out?

PR #295, opened 2021-12-06
Set secure flag for remember_me cookie

paderinandrey:
Added the ability to set the secure flag for the remember_me cookie

- [ ] Description of changes

### DIFF
--- a/lib/sorcery/controller/submodules/remember_me.rb
+++ b/lib/sorcery/controller/submodules/remember_me.rb
@@ -10,9 +10,9 @@ module Sorcery
           base.send(:include, InstanceMethods)
           Config.module_eval do
             class << self
-              attr_accessor :remember_me_httponly
+              attr_accessor :remember_me_httponly, :remember_me_secure
               def merge_remember_me_defaults!
-                @defaults.merge!(:@remember_me_httponly => true)
+                @defaults.merge!(:@remember_me_httponly => true, :@remember_me_secure => false)
               end
             end
             merge_remember_me_defaults!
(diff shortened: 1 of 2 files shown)


athix:
Ah, so this allows users to set secure cookies to true without needing `force_ssl = true`.

Just to sanity check, @paderinandrey does your use-case prevent you from using `force_ssl` (which also enables secure cookies)?

lib/sorcery/controller/submodules/remember_me.rb
```diff
@@ -71,7 +71,8 @@ def set_remember_me_cookie!(user)
               value: user.send(user.sorcery_config.remember_me_token_attribute_name),
               expires: user.send(user.sorcery_config.remember_me_token_expires_at_attribute_name),
               httponly: Config.remember_me_httponly,
-              domain: Config.cookie_domain
+              domain: Config.cookie_domain,
+              secure: Config.remember_me_secure
```
athix, line 75: I'd be worried that this has the potential to overwrite the default of `secure: true` when using `force_ssl`. We'll need to double check that interaction before merging this in. Ideally, adding a test or two.

lib/sorcery/controller/submodules/remember_me.rb
```diff
@@ -10,9 +10,9 @@ def self.included(base)
           base.send(:include, InstanceMethods)
           Config.module_eval do
             class << self
-              attr_accessor :remember_me_httponly
+              attr_accessor :remember_me_httponly, :remember_me_secure
               def merge_remember_me_defaults!
-                @defaults.merge!(:@remember_me_httponly => true)
+                @defaults.merge!(:@remember_me_httponly => true, :@remember_me_secure => false)
```
athix, line 15: `nil` would likely be a safer default here. Still would need to sanity check how this interacts with `force_ssl` either way.

athix (changes requested):
This change is potentially dangerous, depending on how Rails handles the interaction between `force_ssl` and `secure: false`. Needs verification.

willnet:
As @joshbuker  wrote, I think there is no problem if you add `config.force_ssl = true`, so I will close this PR. If there is another use case, please reopen it.

PR #304, opened 2022-02-20
Fix/session timeout

### LINKED ISSUE #303: Session timeout module does not logout user
I have configured sorcery to include session time out module, but when the sorcery try to reset_sessions in validate_session method in controller/submodules/session_timeout.rb the user is not logged out.
Here my config:

config/sorcery.rb
```ruby
  Rails.application.config.sorcery.submodules = [:remember_me, :reset_password, :user_activation, :session_timeout]

  # -- session timeout --
  # How long in seconds to keep the session alive.
  # Default: `3600`
  #
  config.session_timeout = 5 # five second for tests

  # Use the last action as the beginning of session timeout.
  # Default: `false`
  #
  config.session_timeout_from_last_action = false

  # Invalidate active sessions. Requires an `invalidate_sessions_before` timestamp column
  # Default: `false`
  #
  config.session_timeout_invalidate_active_sessions_enabled = true
```

migrate
```ruby
class AddInvalidateSessionToUser < ActiveRecord::Migration[6.1]
  def change
    add_column :users, :invalidate_sessions_before, :datetime
  end
end

```
### Configuration

- Sorcery Version: `0.16.1`
- Ruby Version: `2.7.2`
- Framework: `Rails 6.1.4.4`
- Platform: `Linux`

### Expected Behavior

After 5 seconds of login the session is invalid

### Actual Behavior

The user can still authenticate in platform

### Steps to Reproduce

Just login and wait for 5 seconds, and does not works

### Debug

When i add logout line in sorcery code, its works
```ruby
if (session_to_use && sorcery_session_expired?(session_to_use.to_time)) || sorcery_session_invalidated?
  reset_sorcery_session
  remove_instance_variable :@current_user if defined? @current_user
  logout
else
```

PedroAugustoRamalhoDuarte:

- [ ] Description of changes

# Description
When :remember_me, :session_timeout module are on, the user is not logged out when the first session ends, because session_timeout invalidate the session and remember_me module creates another after that. With this problem we cant logged out a user with timeout when remember_me is on

# Solution
Forget remember me token in session_timeout

closes #303

### DIFF
--- a/lib/sorcery/controller/submodules/session_timeout.rb
+++ b/lib/sorcery/controller/submodules/session_timeout.rb
@@ -52,6 +52,7 @@ module Sorcery
           def validate_session
             session_to_use = Config.session_timeout_from_last_action ? session[:last_action_time] : session[:login_time]
             if (session_to_use && sorcery_session_expired?(session_to_use.to_time)) || sorcery_session_invalidated?
+              forget_me! if current_user.present? && Config.submodules.include?(:remember_me)
               reset_sorcery_session
               remove_instance_variable :@current_user if defined? @current_user
             else
--- a/spec/controllers/controller_session_timeout_spec.rb
+++ b/spec/controllers/controller_session_timeout_spec.rb
@@ -1,7 +1,7 @@
 require 'spec_helper'
 
 describe SorceryController, type: :controller do
-  let!(:user) { double('user', id: 42) }
+  let!(:user) { double('user', id: 42, password: 'secret') }
 
   # ----------------- SESSION TIMEOUT -----------------------
   context 'with session timeout features' do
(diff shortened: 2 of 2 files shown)


athix:
Sorry for the slow response times, been slammed/stressed lately. Haven't forgotten about this, but haven't had the mental bandwidth to tackle it either. Will hopefully get this fixed and tackle the other open issues in the near future, once I'm caught up on other things.

joshbuker:
10 months... Yikes, my bad.

Yeah, adding this to my priority task list.

spec/controllers/controller_session_timeout_spec.rb
```diff
@@ -36,6 +36,61 @@
...
+
+        Timecop.travel(Time.now.in_time_zone + 0.6)
+
+        get :test_login_from_cookie
+
+        expect(cookies['remember_me_token']).to be_nil
+        expect(session[:user_id]).to be_nil
+      end
+
+      it 'does not resets session and remember_me cookies after session timeout' do
```
joshbuker, line 78: Is this test supposed to be "does not reset session and remember_me cookies before session timeout"?

joshbuker (changes requested):
It appears the second test is incorrectly labeled. Can you update it to "remember_me functions before session timeout" or similar, assuming that matches your intention?

joshbuker:
Thanks for your patience :heart: 

Depressingly, even my priority task list is hilariously backlogged.

spec/controllers/controller_session_timeout_spec.rb
PedroAugustoRamalhoDuarte: Yes, i will change

PedroAugustoRamalhoDuarte:
Thank you for the review, I will improve this specs next week and test a little bit more. This line ` Timecop.travel(Time.now.in_time_zone + 0.6)` looks like magic, i will try to explicit setup the session timeout for make the specs easier to read

lib/sorcery/controller/submodules/session_timeout.rb
```diff
@@ -52,6 +52,7 @@ def register_login_time(_user, _credentials = nil)
           def validate_session
             session_to_use = Config.session_timeout_from_last_action ? session[:last_action_time] : session[:login_time]
             if (session_to_use && sorcery_session_expired?(session_to_use.to_time)) || sorcery_session_invalidated?
+              forget_me! if current_user.present? && Config.submodules.include?(:remember_me)
```
jankoegel, line 55: Wouldn't it make more sense to just call `logout` here?
There might be other before/after logout hooks configured and `logout` already takes care of calling all of those.
Just discovered this PR after I ran into the same problem and opened my PR with `logout` here:
https://github.com/Sorcery/sorcery/pull/358

PedroAugustoRamalhoDuarte: Hello @jankoegel i saw you PR, yes is better then mine. Makes more sense to call logout. Can i add the tests in your PR?

jankoegel: Sure, I invited you as a collaborator to my fork. Let me know if that works for you & thanks for preparing the specs!

PedroAugustoRamalhoDuarte:
@joshbuker @jankoegel closing to resolve in https://github.com/Sorcery/sorcery/pull/358

PR #307, opened 2022-02-23
Fix broken builds (net/smtp)

athix:
Currently the builds are broken due to changes in how net/smtp is included in ruby itself. Updating the version matrices and gemfiles to address these changes.


athix:
Hmm. Seems like Rails 7 is broken due to issues with session timeout. I'll remove Rails 7 for the matrix just so I can get the builds passing, but I would like to officially get Rails 7 working sooner rather than later.

PR #315, opened 2022-06-30
Added support for `scope_for_authentication`  method

iamdeuterium:
Example:
```
class User < ApplicationRecord
  authenticates_with_sorcery!

  def self.scope_for_authentication
    where(deleted: false)
  end
end
```


KitaitiMakoto:
Until this pull request is merged, there is a workaround:

### config/initializers/sorcery.rb ###
```config/initializers/sorcery.rb
Rails.application.config.sorcery.configure do |config
  #  :
  #  :
  config.user_class = 'ActiveUser'
  #  :
  #  :
end
```

### app/models/active_user.rb ###
```app/models/active_user.rb
ActiveUser = User.where(deleted: false)
```

PR #332, opened 2023-01-18
Implement #login! helper

### LINKED ISSUE #327: Feature request: `login!` method
I'd like to suggest adding a `login!` method, that would behave like `login` except it would raise an exception if the login fails.

This is similar to the `save` / `save!` methods in Rails.

It would be helpful in cases where you want to code the "happy path", and handle exceptions separately. For instance instead of:

```rb
class Api::V1::SessionsController < ApplicationController
  def create
    if login(params[:email], params[:password])
      render json: { message: "You're logged in!" }
    else
      render json: { message: "Wrong credentials" }
    end
  end
end
```

you could have:

```rb
class Api::V1::SessionsController < ApplicationController
  rescue_from Sorcery::InvalidCredentials, with: :invalid_credentials

  def create
    login!(params[:email], params[:password])
    
    render json: { message: "You're logged in!" }
  end

  private

  def invalid_credentials
    render json: { message: "Wrong credentials" }
  end
end
```

A few notes:

- I used `render json` for the sake of simplicity but view templates would be used instead
- `Sorcery::InvalidCredentials` is just a suggestion, I'm open to ideas for a better name

Spone:
Close #327

This PR adds a `login!` method, that behaves like `login` except it raises an exception if the login fails.
See #327 for more details.

- [ ] Description of changes


lib/sorcery/controller.rb
```diff
@@ -1,4 +1,6 @@
 module Sorcery
+  class InvalidCredentials < StandardError; end
```
Spone, line 2: Not sure if it's the best place to define this exception?

joshbuker: Here's what I did in the v1 codebase:

https://github.com/Sorcery/sorcery-rework/blob/main/sorcery-core/lib/sorcery/errors.rb

If you follow a similar pattern, it should be easy to migrate this functionality into the new codebase, once I finally can sit down and finish it.

lib/sorcery/controller.rb
```diff
@@ -63,6 +65,16 @@ def login(*credentials)
...
       end
 
+      def login!(*credentials)
+        user = login(*credentials)
+
+        if user.nil?
+          raise Sorcery::InvalidCredentials
+        else
+          user
+        end
```
joshbuker, line 75:
```ruby
user = login(*credentials)

raise Sorcery::Error::InvalidCredentials if user.nil?

user
```

I suspect RuboCop will bark about the spacing between lines, but this will at least use a guard clause instead of an if/else.

lib/sorcery/controller.rb
```diff
@@ -63,6 +65,16 @@ def login(*credentials)
         end
       end
 
+      def login!(*credentials)
```
joshbuker, line 68: We'll need to test this and see how it interacts when given a code block to yield to.

spec/controllers/controller_spec.rb
```diff
@@ -85,6 +85,39 @@
       end
     end
 
+    describe '#login!' do
+      context 'when succeeds' do
+        before do
+          expect(User).to receive(:authenticate).with('bla@bla.com', 'secret') { |&block| block.call(user, nil) }
```
joshbuker, line 91: It looks like this test calls User.authenticate directly rather than the login method. The controller spec should test the controller methods.

joshbuker: Ah, wait, never mind. It's been a minute since I've looked at Ruby. This is overwriting the authenticate method for User during the duration of the test.

Is this necessary?

joshbuker (changes requested):
Please update the error class to mirror the sorcery/errors.rb in v1. That would help with porting this forward into the v1 codebase.

I'll need to take a closer look at the tests and run them locally to understand better what's going on there. It might be fine as-is, but as the login method is critical to the security of Sorcery, I want to be very thorough with testing its functionality.

spec/controllers/controller_spec.rb
Spone: I duplicated this code from the `#login` tests [above](https://github.com/Sorcery/sorcery/pull/332/files/9ebaf3166c99b63aa2f3c14dc48a4f4c5b537387#diff-1f60346fac30942dff621cab795a97495a4117679cec4ba682995891a0628e3cL55), but i can remove it if it's unnecessary.

lib/sorcery/controller.rb
Spone: Done in 0592c84

lib/sorcery/controller.rb
Spone: I'm not sure what the behavior should be when this method is called with a block. What do you suggest?

spec/controllers/controller_spec.rb
joshbuker: :facepalm: 

That's fine, then. I think it would be better to be consistent and fix it all at once later.

lib/sorcery/controller.rb
joshbuker: Essentially, sanity checks that the method performs as expected when passing a block.

This wiki page has an example of using the block for raising different error messages during login: https://github.com/Sorcery/sorcery/wiki/Distinguish-login-failure-reasons

When given a code block, we can test that it succeeds when given valid credentials, fails when given invalid credentials, and raises your new error when given a blank string or nil value.

joshbuker (dismissed):
This needs some test cases for when `login!` is passed a code block, but otherwise looks good!

lib/sorcery/controller.rb
Spone: @joshbuker I added specs for the block version of `#login!`. I have a few notes:
- I can't find tests for the block version of `#login`
- I'm not sure what the behavior should be when calling `#login!` with a block and with wrong credentials. I would expect it to always raise the `InvalidCredentials` exception, else I don't see the value in using `#login!`

Spone:
Just fixed conflicts, if you want to review again @joshbuker

Spone:
@willnet sorry I won't have time to dedicate to this in the coming weeks

willnet (changes requested):

lib/sorcery/controller.rb
```diff
@@ -57,6 +57,14 @@ def login(*credentials)
         end
       end
 
+      def login!(*credentials, &)
+        user = login(*credentials, &)
+
+        raise Sorcery::InvalidCredentials if user.nil?
+
+        user
```
joshbuker: LGTM

lib/sorcery.rb
```diff
@@ -1,4 +1,5 @@
 require 'sorcery/version'
+require 'sorcery/errors'
```
joshbuker: :rocket:

PR #333, opened 2023-01-18
Raise ArgumentError when calling change_password! with blank password

### LINKED ISSUE #329: `#change_password!` with an empty string fails silently
### Configuration

- Sorcery Version: `0.16.4`
- Ruby Version: `3.1.2`
- Framework: `Rails 7.0.4`
- Platform: `Linux`

### Expected Behavior

When I'm calling `@user.change_password!("")`, I would expect an exception to be raised, because the password is missing, so it won't be modified.

### Actual Behavior

The method does not raise an exception and returns `true`.

### Steps to Reproduce

In Rails console:

1. Create a `User`
2. Call `User.last.change_password!("")`
3. No exception is raised and `true` is returned

Do you consider this a bug? If yes, I can provide a test case in a PR.

Spone:
Close #329

When calling `@user.change_password!("")` or `@user.change_password!(nil)`, a `ArgumentError` exception is now raised, because the password is missing, so it won't be modified.

- [x] Description of changes
- [x] Update to CHANGELOG.md with short description and link to pull request

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 ## HEAD
 
+* Raise ArgumentError when calling change_password! with blank password [#333](https://github.com/Sorcery/sorcery/pull/333)
+
 ## 0.16.4
 
 * Adapt to open request protection strategy of rails 7.0 [#318](https://github.com/Sorcery/sorcery/pull/318)
--- a/lib/sorcery/model/submodules/reset_password.rb
+++ b/lib/sorcery/model/submodules/reset_password.rb
@@ -131,6 +131,8 @@ module Sorcery
           end
 
           def change_password!(new_password)
+            raise ArgumentError, 'Blank password passed to change_password!' if new_password.blank?
+
             change_password(new_password, raise_on_failure: true)
           end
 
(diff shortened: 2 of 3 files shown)


joshbuker:
Digging through the code, I can't put the finger on where it prevents change_password from accepting a nil/blank password. As far as I can tell, it should pass it to the crypto provider (usually bcrypt), which will accept nil/blank. That said, I tried it locally, and change_password will indeed do nothing if you set the new password as nil/blank.

Not necessarily a blocker on this PR, but I would like to understand better what is going on there.

Spone:
I only changed the bang version (`change_password!`). Would it make sense to modify `change_password`'s behavior as well? I'm not sure.

Spone:
@joshbuker I'm awaiting further instructions here :)

joshbuker:
At the dentist, I'll take another look and probably merge this in about an hour.

joshbuker (approved):

joshbuker:
Looks good, at some point I should dig more into that code to understand what's going on, and this is a bit of a breaking change API-wise. That said, I suspect it shouldn't be an issue, considering most (all?) applications would raise an error on blank password with the virtual attribute validators anyway.

PR #338, opened 2023-03-05
Fix timestamped_migrations method call since deprecated

phChun:
Per Rails 7.x, 
```
DEPRECATION WARNING: ActiveRecord::Base.timestamped_migrations is deprecated and will be removed in Rails 7.1.
Use `ActiveRecord.timestamped_migrations` instead.   
```

This method was removed in Rails 7.1, breaking the sorcery installer. Following the guidance from the deprecation warning, the fix listed above works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 ## HEAD
-
+* Fix deprecated call to timestamped_migrations in the install generateor [#338](https://github.com/Sorcery/sorcery/pull/338)
 * Raise ArgumentError when calling change_password! with blank password [#333](https://github.com/Sorcery/sorcery/pull/333)
 
 ## 0.16.4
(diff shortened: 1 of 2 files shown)


lib/generators/sorcery/install_generator.rb
```diff
@@ -75,7 +75,7 @@ def copy_migration_files
 
       # Define the next_migration_number method (necessary for the migration_template method to work)
       def self.next_migration_number(dirname)
-        if ActiveRecord::Base.timestamped_migrations
+        if ActiveRecord.timestamped_migrations
```
joshbuker, line 78: Is this backwards compatible with Rails 6? That would be my only concern merging this in.

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Changelog
 ## HEAD
-
+* Fix deprecated call to timestamped_migrations in the install generateor [#338](https://github.com/Sorcery/sorcery/pull/338)
```
joshbuker, line 3: Small typo, `generateor` -> `generator`

joshbuker (changes requested):
IIRC, there aren't any tests currently for the generators, so it's not guaranteed that this doesn't break backwards compatibility with Rails 6.

If you could verify that it still works when using Rails 6, I can go ahead and merge this in.

PR #339, opened 2023-03-15
Remove duplicate method in InstallGenerator

### LINKED ISSUE #265: Install generator gives an error
### Configuration

- Sorcery Version: `master`
- Ruby Version: `2.7.2`
- Framework: `Rails 6.1`
- Platform: `Debian-slim (docker)`

### Expected Behavior

No errors during generation of new Sorcery setup.

### Actual Behavior

When running `rails generate sorcery:install --model=Admin::User` i get the following:
```bash
      create  config/initializers/sorcery.rb
    generate  model Admin::User --skip-migration
       rails  generate model Admin::User --skip-migration
Running via Spring preloader in process 419
      invoke  active_record
      create    app/models/admin/user.rb
      create    app/models/admin.rb
      invoke    test_unit
      create      test/models/admin/user_test.rb
      create      test/fixtures/admin/users.yml
      insert  app/models/admin/user.rb
File unchanged! The supplied flag value not found!  app/models/admin/user.rb
      create  db/migrate/20201201102702_sorcery_core.rb
```

### Steps to Reproduce

`rails generate sorcery:install --model=Admin::User`

### Additional comments

The odd thing is is that the resulting model looks correct:

```ruby
class Admin::User < ApplicationRecord
  authenticates_with_sorcery!
end
```
I've made a quick peek at the generator and i can't seem to find where the message comes from.

P.s. is there any info on a new release? 
I'd rather not point to the git repo and the master contains some useful fixes.

Spone:
There's no need to call `inject_sorcery_to_model` in the previous method, since the generators execute all public methods in order.

That's why we had an error in the logs: the method was called twice.

Close #265

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 ## HEAD
 
+* Fix error when running the install generator [#339](https://github.com/Sorcery/sorcery/pull/339)
 * Raise ArgumentError when calling change_password! with blank password [#333](https://github.com/Sorcery/sorcery/pull/333)
 
 ## 0.16.4
--- a/lib/generators/sorcery/install_generator.rb
+++ b/lib/generators/sorcery/install_generator.rb
@@ -48,7 +48,6 @@ module Sorcery
         return if only_submodules?
 
         generate "model #{model_class_name} --skip-migration"
-        inject_sorcery_to_model
       end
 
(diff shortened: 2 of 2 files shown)


Spone:
I fixed the conflict with `master. @joshbuker would you mind reviewing this?

joshbuker (approved):

PR #318, opened 2022-07-20
adapt to Rails 7.0 defaults

craisp:
Since 7.0, Rails protects against redirecting to external hosts
https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_to-label-Open+Redirect+protection
This change adapts to this by adding the option `allow_other_host` to the method call

PR #351, opened 2023-10-26
Allow open redirect for Rails 7.0 in #redirect_back_or_to method

berkos:
👋  Thanks for the review in advance.
Fixes open redirect for Rails 7.0
https://api.rubyonrails.org/classes/ActionController/Redirecting.html#method-i-redirect_to-label-Open+Redirect+protection

The PR here follows a similar approach with https://github.com/Sorcery/sorcery/pull/318

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 ## HEAD
 
 * Allow open redirect for Rails 7.0 in #redirect_back_or_to method [#351](https://github.com/Sorcery/sorcery/pull/351)
+
+## 0.17.0
+
 * Fix Rails 7.1 compatibility by using `ActiveRecord.timestamped_migrations` [#352](https://github.com/Sorcery/sorcery/pull/352)
 * Change CI settings for support Ruby3.0+ Rails6.1+ [#357](https://github.com/Sorcery/sorcery/pull/357)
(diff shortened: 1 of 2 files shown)


lib/sorcery/controller.rb
```diff
@@ -98,7 +98,7 @@ def current_user=(user)
       # used when a user tries to access a page while logged out, is asked to login,
       # and we want to return him back to the page he originally wanted.
       def redirect_back_or_to(url, flash_hash = {})
-        redirect_to(session[:return_to_url] || url, flash: flash_hash)
+        redirect_to(session[:return_to_url] || url, flash: flash_hash, allow_other_host: true)
```
willnet, line 101: The redirect_back_or_to method, unlike what's described in https://github.com/Sorcery/sorcery/pull/318, might accept user input depending on how it's used. Therefore, it would be more secure and preferable to not set allow_other_host to true by default, but rather, to provide it as an argument.

lib/sorcery/controller.rb
```diff
@@ -98,7 +98,7 @@ def current_user=(user)
       # used when a user tries to access a page while logged out, is asked to login,
       # and we want to return him back to the page he originally wanted.
       def redirect_back_or_to(url, flash_hash = {})
```
dapi, line 100: Add here `allow_other_host: false` and use it next lines

willnet:
Addressed in #406

Issue #330, opened 2022-11-28
Using sorcery in an API-only rails app does not work


TheNeikos:
### Configuration

- Sorcery Version: `0.16.4`
- Ruby Version: `3.1.2`
- Framework: `rails-7`
- Platform: `linux-64_86`

### Expected Behavior

Sorcery should work out of the box with an API-only rails app

### Actual Behavior

It errors with `undefined local variable or method 'form_authenticity_token'`.

### Steps to Reproduce

1. Make a rails API app
2. Install sorcery
3. Call `login`

mladenilic:
Login method calls `form_authenticity_token` which is defined in request forgery protection concern. You can include it with:
```
include ActionController::RequestForgeryProtection
```

Nitrodist:
@mladenilic saved me some work here, thanks!

PR #150, opened 2018-09-25
Fix rails injection

tiii:
Fixes #137.

This PR correctly loads sorcery.

PR #357, opened 2024-01-31
Change CI settings for support Ruby3.0+ Rails6.1+ 

### LINKED ISSUE #340: Deprecate Ruby 2.7
Per https://github.com/ossf/osv-schema/issues/123#issuecomment-1490334983, Ruby 2.7 has reached end-of-life, and the new minimum version should be at least 3.0 which is in security-only maintenance.

Need to update the CI and gemspec accordingly.

willnet:
Now, Ruby 2.7 is EOL, and Rails 6.0 is also EOL.

So, I changed the CI settings to support only those higher versions and made the changes necessary to pass the tests.

CI result is [here](https://github.com/willnet/sorcery/actions/runs/7724024577).

Fix #340

### DIFF
--- a/gemfiles/rails_61.gemfile
+++ b/gemfiles/rails_61.gemfile
@@ -3,5 +3,5 @@ source 'https://rubygems.org'
 gem 'rails', '~> 6.1.0'
 gem 'rails-controller-testing'
 gem 'sqlite3', '~> 1.4'
-
+gem 'rspec-rails', '>= 5.0'
 gemspec path: '..'
--- a/sorcery.gemspec
+++ b/sorcery.gemspec
@@ -38,7 +38,6 @@ Gem::Specification.new do |s|
   s.add_dependency 'oauth2', '~> 2.0'
 
   s.add_development_dependency 'byebug', '~> 10.0.0'
-  s.add_development_dependency 'rspec-rails', '~> 3.7.0'
   s.add_development_dependency 'rubocop'
   s.add_development_dependency 'simplecov', '>= 0.3.8'
   s.add_development_dependency 'test-unit', '~> 3.2.0'
--- a/sorcery.gemspec
+++ b/sorcery.gemspec
@@ -38,6 +38,7 @@ Gem::Specification.new do |s|
   s.add_dependency 'oauth2', '~> 2.0'
 
   s.add_development_dependency 'byebug', '~> 10.0.0'
+  s.add_development_dependency 'rspec-rails'
   s.add_development_dependency 'rubocop'
   s.add_development_dependency 'simplecov', '>= 0.3.8'
   s.add_development_dependency 'test-unit', '~> 3.2.0'
--- a/spec/controllers/controller_session_timeout_spec.rb
+++ b/spec/controllers/controller_session_timeout_spec.rb
@@ -129,9 +129,11 @@ describe SorceryController, type: :controller do
     end
 
     context "with 'session_timeout_from_last_action'" do
+      before { create_new_user }
+      after { User.delete_all }
+
       it 'does not logout if there was activity' do
         sorcery_controller_property_set(:session_timeout_from_last_action, true)
-        expect(User).to receive(:authenticate).with('bla@bla.com', 'secret') { |&block| block.call(user, nil) }
 
         get :test_login, params: { email: 'bla@bla.com', password: 'secret' }
         Timecop.travel(Time.now.in_time_zone + 0.3)
--- a/spec/controllers/controller_spec.rb
+++ b/spec/controllers/controller_spec.rb
@@ -64,7 +64,13 @@ describe SorceryController, type: :controller do
           expect(session[:user_id]).to eq user.id.to_s
         end
 
+        # NOTE: The lack of a CSRF token may mean that sessions will break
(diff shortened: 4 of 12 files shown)


gemfiles/rails_61.gemfile
```diff
@@ -3,5 +3,5 @@ source 'https://rubygems.org'
 gem 'rails', '~> 6.1.0'
 gem 'rails-controller-testing'
 gem 'sqlite3', '~> 1.4'
-
+gem 'rspec-rails', '>= 5.0'
```
joshbuker, line 6: >= version locks can cause issues, but I'm not concerned enough about it to block the PR.

sorcery.gemspec
```diff
@@ -38,7 +38,6 @@ Gem::Specification.new do |s|
   s.add_dependency 'oauth2', '~> 2.0'
 
   s.add_development_dependency 'byebug', '~> 10.0.0'
-  s.add_development_dependency 'rspec-rails', '~> 3.7.0'
```
joshbuker, line 41: Might be better to just remove the version lock instead, but I vaguely recall dev dependencies in the gemspec being deprecated anyway.

spec/controllers/controller_session_timeout_spec.rb
```diff
@@ -129,9 +129,11 @@
     end
 
     context "with 'session_timeout_from_last_action'" do
+      before { create_new_user }
+      after { User.delete_all }
+
       it 'does not logout if there was activity' do
         sorcery_controller_property_set(:session_timeout_from_last_action, true)
-        expect(User).to receive(:authenticate).with('bla@bla.com', 'secret') { |&block| block.call(user, nil) }
```
joshbuker, line 134: Because this touches authentication, I'll need to verify before merging this.

spec/controllers/controller_spec.rb
```diff
@@ -63,10 +63,6 @@
         it 'writes user id in session' do
           expect(session[:user_id]).to eq user.id.to_s
         end
-
-        it 'sets csrf token in session' do
-          expect(session[:_csrf_token]).not_to be_nil
-        end
```
joshbuker, line 69: Because this touches csrf, I'll need to verify before merging this.

joshbuker (commented):
Need to verify before merging.

joshbuker (approved):
NGL, I can't find any mentions of changes to the csrf functionality that would cause the test to fail in Rails 7.1. Has me pretty spooked, but I don't have the time to investigate further and Rails 7.1 is largely broken in its current state for Sorcery v0 anyway.

Really need to finish that v1 rework, but it's been long enough that getting back into it is challenging.

spec/controllers/controller_spec.rb
```diff
@@ -64,7 +64,13 @@
           expect(session[:user_id]).to eq user.id.to_s
         end
 
+        # NOTE: The lack of a CSRF token may mean that sessions will break
+        #       horribly for Sorcery when using Rails 7.1+. We shall see.
         it 'sets csrf token in session' do
+          if Gem::Version.new(Rails.version) >= Gem::Version.new('7.1')
+            pending 'Rails 7.1 is not including the csrf token in the session for unknown reasons'
+          end
```
willnet, line 72: @joshbuker Sorcery uses the [form_authenticity_token method](https://github.com/Sorcery/sorcery/blob/856eacd9912ecbb436a7259ba927f171d77b0ba1/lib/sorcery/controller.rb#L57) to assign values to `session[:_csrf_token]`. However, it seems that due to [this commit](https://github.com/rails/rails/commit/f2c66ce392016efd49816cd6fdc9f174271a0321) in Rails 7.1, the behavior of the form_authenticity_token method has changed and it no longer assigns values to `session[:_csrf_token]`.

If it's necessary to assign a value to `session[:_csrf_token]` during the execution of the login method, then we may need to revise our implementation. The question arises whether this is actually necessary. Rails seems to assign values to `session[:_csrf_token]` as needed, so there might not be a clear need for Sorcery to explicitly do this. There doesn't seem to be a specific reason mentioned in the commit where [form_authenticity_token was introduced](https://github.com/Sorcery/sorcery/commit/8a7490c3b6b8e64c9b6d092103a74be77e4093e6), making it hard to infer the rationale behind this implementation.

willnet: @joshbuker I did some additional research on this topic. The PR that introduced the form_authenticity_token method is [here](https://github.com/NoamB/sorcery/pull/409), and it seems it was merged to fix [this issue](https://github.com/NoamB/sorcery/issues/250) (though it's mentioned as `Fix for #150` in the PR, it seems like a typo and should probably be `Fix for #250` based on the content).

However, calling form_authenticity_token during login doesn't resolve the issue mentioned in https://github.com/NoamB/sorcery/issues/250, making [this commit](https://github.com/Sorcery/sorcery/commit/8a7490c3b6b8e64c9b6d092103a74be77e4093e6) seem rather pointless. Moreover, invoking form_authenticity_token has led to a new issue being created, as seen in [Using sorcery in an API-only rails app does not work · Issue #330 · Sorcery/sorcery](https://github.com/Sorcery/sorcery/issues/330), suggesting that it might be better to remove this implementation and related tests.

PR #370, opened 2024-04-03
Remove legacy Rails version conditionals

willnet:
The current sorcery now supports Rails >= 6.1 and Ruby >= 3.0. I've removed and cleaned up the code that was branching for older versions of Rails.


lib/generators/sorcery/install_generator.rb
```diff
@@ -97,11 +97,7 @@ def only_submodules?
       end
 
       def migration_class_name
-        if Rails::VERSION::MAJOR >= 5
-          "ActiveRecord::Migration[#{Rails::VERSION::MAJOR}.#{Rails::VERSION::MINOR}]"
-        else
-          'ActiveRecord::Migration'
-        end
+        "ActiveRecord::Migration[#{Rails::VERSION::MAJOR}.#{Rails::VERSION::MINOR}]"
```
joshbuker, line 100: I think we might be able to refactor this to just use `ActiveRecord::Migration::Current`, but non-blocking.

joshbuker (approved):
This is great, thanks @willnet!

lib/generators/sorcery/install_generator.rb
willnet: @joshbuker When using ActiveRecord::Migration::Current here, the content of the migration file might change when users upgrade their Rails version, so I think it's best to leave it as is.

joshbuker: Wouldn't that also be true using this method?

PR #371, opened 2024-04-10
Remove form_authenticity_token method

### LINKED ISSUE #330: Using sorcery in an API-only rails app does not work

willnet:
Fix #330

As detailed in the URL below, the call to `form_authenticity_token` wasn't useful and was causing issues like https://github.com/Sorcery/sorcery/issues/330, so remove it.

ref:
- https://github.com/Sorcery/sorcery/pull/357#discussion_r1522888757
- https://github.com/Sorcery/sorcery/pull/357#discussion_r1559150755


Spone:
Why wasn't is useful?

What about including the module as described in https://github.com/Sorcery/sorcery/issues/330#issuecomment-1329540334?

joshbuker:
@Spone to invert the question, what is the purpose of calling `form_authenticity_token`? That's something I don't fully understand yet personally, and from what @willnet is suggesting, it's actually not doing anything useful.

My assumption would be that it's to generate the csrf token in a scenario where rails wouldn't already do so, but that seems like it might be a legacy issue that is no longer relevant.

Spone:
Apparently it was added in https://github.com/Sorcery/sorcery/commit/8a7490c3b6b8e64c9b6d092103a74be77e4093e6 in order to add `_csrf_token` to the session.

I agree that if the behavior is not fully understood, and that removing the line does not break any test, and furthermore helps solve a current issue, we should do it.

Spone (approved):

willnet:
As I mentioned [here](https://github.com/Sorcery/sorcery/pull/357#discussion_r1559150755), I don't have enough information from [this PR](https://github.com/NoamB/sorcery/pull/409) to be certain, but I believe form_authenticity_token was added in an attempt to solve [this issue](https://github.com/NoamB/sorcery/issues/250). My guess is that it was meant to maintain `session[:_csrf_token]` even after reset_session is executed. However, in reality, `session[:_csrf_token]` does not remain but is updated, so if my guess is correct, this method call becomes particularly meaningless.

willnet:
I have resolved the conflict in CHANGELOG.md

joshbuker (approved):

Issue #296, opened 2021-12-31
Should Sorcery rename redirect_back_or_to?


martinzamuner:
Rails 7 released a new method called `redirect_back_or_to` added by DHH as a replacement for the now soft-deprecated `redirect_back`: https://github.com/rails/rails/pull/40671.

That may conflict with the method by the same name defined by Sorcery. Should it be renamed? I'm opening an issue instead of a PR because I guess it's not a trivial decision to make.

athix:
Hmm. Yeah, that'll require some thought. The most robust solution would probably be to namespace the method, but that wouldn't be backwards compatible. Could dynamically define it based on what version of rails is defined, but I prefer avoiding that when possible just because it introduces non-trivial amounts of complexity.

Thanks for bringing this up. Rails 7 has a lot of nice additions, so I want to support it as soon as possible.

I'm currently knee deep in both a personal and work project, but I'll tackle Rails 7 myself after those are both settled. In the meantime, any Rails 7 related issues that can be fixed in a backwards compatible way I'll do my best to merge in quickly.

lorenzk:
FYI I am using this workaround for the time being:

```ruby
# config/initializers/sorcery_workaround_for_redirect_back_or_to.rb
module Sorcery
  module Controller
    module InstanceMethods
      define_method :sorcery_redirect_back_or_to, instance_method(:redirect_back_or_to)
      remove_method :redirect_back_or_to
    end
  end
end
```

And then use `sorcery_redirect_back_or_to` in my controllers when I want to use the `session[:return_to]` feature and rails' `redirect_back_or_to` in place of the soft-deprecated `redirect_back`.

Maybe sorcery could hook into rails' mechanism by just favoring `session[:return_to]` over `request.referer` over `fallback_location`?

Exterm1nate:
A good way to solve this is to define another method with same functionality (something like `redirect_to_before_login_path`), add a warning message to the original method that it is deprecated and will be soon removed, and finally don't define it at all for `rails >= 7`, because it breaks the functionality of passed options (they are passed to the flash kwargs).

kleinjm:
Bump 🙏  I went down a bit of a debugging rabbit hole due to this naming conflict

PedroAugustoRamalhoDuarte:
I got this error in production today, strange that in my development workspace is calling the rails redirect_back_or_to in and in production the sorcery one. Is there anyone looking to handle this problem?

joshbuker:
[...]

PR #373, opened 2024-05-01
Add an option 'use_redirect_back_or_to_by_rails' to avoid definition conflicts of `redirect_back_or_to`

atolix:
Fix: https://github.com/Sorcery/sorcery/issues/296

Rails 7 released a new method called `redirect_back_or_to` as a replacement for `redirect_back`. That may conflicts with the method by the same name defined by Sorcery.

This commit adds an option to set whether to use `redirect_back_or_to` defined by Rails 7, and add the method `redirect_to_before_login_path` as an alternative to sorcery's `redirect_back_or_to.

ref: https://github.com/rails/rails/pull/40671

- [x] Description of changes


lib/sorcery/controller.rb
```diff
@@ -96,7 +96,16 @@ def current_user=(user)
 
       # used when a user tries to access a page while logged out, is asked to login,
       # and we want to return him back to the page he originally wanted.
-      def redirect_back_or_to(url, flash_hash = {})
+      def redirect_back_or_to(...)
+        if Config.use_redirect_back_or_to_by_rails
```
willnet, line 100: If you are using Rails 6.1, calling super will cause an exception. Therefore, I want to ensure that super is not called even if Config.use_redirect_back_or_to_by_rails returns true.

lib/sorcery/controller.rb
```diff
@@ -96,7 +96,16 @@ def current_user=(user)
...
+          super
+        else
+          warn('[WARNING] `redirect_back_or_to` overrides the method of the same name defined in Rails 7. If you want to avoid overriding, you can set `config.use_redirect_back_or_to_by_rails = true` and use `redirect_to_before_login_path`.')
```
willnet, line 103: Since it is desirable for this setting to default to true in the future, I would like to add that "in a future version, `config.use_redirect_back_or_to_by_rails = true` will become the default."

willnet (changes requested):

lib/sorcery/controller.rb
atolix: I have changed the message in  https://github.com/Sorcery/sorcery/pull/373/commits/8ebe075834d2954b3a557cf74142babb7869f5e9

willnet (approved):
I'm not very good at English, so I'm not sure if the comments and warning messages sound natural, but I think the implementation is fine.

willnet:
@joshbuker  As far as I have reviewed, I think the implementation is fine, but I'm not sure if the comments and other texts are clear. Could you please review them?

willnet (changes requested):
I've reviewed your PR again.

Could you please change the comment to use `redirect_to_before_login_path` instead of `redirect_back_or_to` in the initializer file ( https://github.com/sorcery/sorcery/blob/b3e7793e44ff649938302c9fd9bdb493179af06c/lib/generators/sorcery/templates/initializer.rb#L19 )?

Also, in the test dummy application's controller ( https://github.com/sorcery/sorcery/blob/b3e7793e44ff649938302c9fd9bdb493179af06c/spec/rails_app/app/controllers/sorcery_controller.rb ), `redirect_back_or_to` is used in actions other than `test_return_to` (which is used by your new test code). Please update these other actions to use `redirect_to_before_login_path`.

atolix:
@willnet 
Thank you for the review, and sorry for the late reply.

>Could you please change the comment to use redirect_to_before_login_path instead of redirect_back_or_to in the initializer file ( https://github.com/sorcery/sorcery/blob/b3e7793e44ff649938302c9fd9bdb493179af06c/lib/generators/sorcery/templates/initializer.rb#L19 )?

fixed in https://github.com/Sorcery/sorcery/pull/373/commits/d78f5fed1eec4fbd66abd4d36f986299dcd9d234

>Also, in the test dummy application's controller ( https://github.com/sorcery/sorcery/blob/b3e7793e44ff649938302c9fd9bdb493179af06c/spec/rails_app/app/controllers/sorcery_controller.rb ), redirect_back_or_to is used in actions other than test_return_to (which is used by your new test code). Please update these other actions to use redirect_to_before_login_path.

fixed in https://github.com/Sorcery/sorcery/pull/373/commits/3077b41590fc529bf1c2fc2c65ef00041c399883

spec/controllers/controller_spec.rb
```diff
@@ -186,5 +192,42 @@
...
+              expect(response).to redirect_to('http://test.host/referer_action')
+            end
+          end
+
+          context 'when Rails::VERSION::MAJOR < 7', skip: Rails::VERSION::MAJOR >= 7 do
+            it 'raise NoMethodError' do
+              expect { get :test_redirect_back_or_to }.to raise_error(NoMethodError)
+            end
+          end
+        end
```
willnet, line 217: [Drop support for versions below Ruby 3.2 and Rails 7.1 (#383) · Sorcery/sorcery@ea120d8](https://github.com/Sorcery/sorcery/commit/ea120d8e00eae5b28c6998c2d25deaa3477161f6) dropped support for Rails < 7.1, so please remove tests that target Rails < 7.

atolix: Removed the test in https://github.com/Sorcery/sorcery/pull/373/commits/377f2ed2194e44bd3bc4a088563b9dd71b8733e8

atolix:
@brendon @willnet 

Thanks for the feedback. Just to clarify, are you suggesting removing Sorcery’s own `redirect_back_or_to` method entirely (and relying only on Rails’ version), or renaming Sorcery’s method instead? 

Either way, I agree with simplifying it. Since it would be a breaking change, I just want to confirm that this is acceptable to move forward with.

brendon:
Yes I think it makes sense to remove the custom `redirect_back_or_to` and rely on the Rails version. I imagine the breaking change would be that users would need to update their use of the method to conform to the Rails version's signature?

I haven't checked, but I assume the purpose of both methods is the same? or is there more nuance to the Sorcery version?

willnet:
I support the approach in this PR to first show a warning.

Sorcery’s redirect_back_or_to prioritizes redirecting to the URL the user tried to access while not logged in (when that URL requires authentication).

https://github.com/Sorcery/sorcery/blob/301ada945d077ec8e8198d546a42edbddf77ea07/lib/sorcery/controller.rb#L99-L102

Rails’ redirect_back_or_to prioritizes redirecting to the referrer URL.

https://github.com/rails/rails/blob/2ea1d9bd035e9f41d535d673efbaf4173441ab34/actionpack/lib/action_controller/metal/redirecting.rb#L199-L207

If we switch from the Sorcery version to the Rails version all at once with no migration period, their argument signatures are almost the same, so no exception will be raised and Sorcery users might not notice. However, the application behavior will change.

Even if we log a message as in the current implementation of this PR, people might miss the logs, but it’s still better than switching abruptly. To ensure users notice this breaking change, we could consider raising an exception in the version after the one that includes this PR when redirect_back_or_to is used without `config.use_redirect_back_or_to_by_rails = true`.

> Given new releases of sorcery aren’t very frequent it might just be better to whip the band-aid off in one go?

Once this PR and #351 are resolved, I’d like to discuss releasing the next version with @joshbuker . If I can get release permissions, I’d like to release more frequently than we do now.

willnet:
Since Sorcery now supports Rails 7.1 and above, we can use Rails.application.deprecators. I think that’s a great idea.

atolix:
I’ve added a test in https://github.com/Sorcery/sorcery/pull/373/commits/1c92b2770f7ae78a798282e294f0a8eb748683ee

brendon (approved):
I think this is ready to merge :)

willnet (approved):

PR #374, opened 2024-07-03
Remove unused SimpleCov

willnet:
I couldn’t find any code using SimpleCov, other than the commented-out sections removed in this commit, so it seems SimpleCov is not currently in use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 ## HEAD
 
+* Remove unused SimpleCov [#374](https://github.com/Sorcery/sorcery/pull/374) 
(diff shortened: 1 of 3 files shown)


joshbuker (approved):
This will be properly utilized in v1, but for now no reason to keep it if it's fully disabled. I think v0 also doesn't have RuboCop enabled either iirc.

willnet:
@joshbuker Thank you for merging. In my opinion, since no one has enough time to complete v1 at the moment, it would be better to gradually improve v0. Therefore, if you think it would be good to have test coverage or RuboCop in sorcery, I can enable the settings. What do you think?

PR #377, opened 2025-01-22
Fixed minor issues with test to get all green so that we can continue development.

Beartech:

- Expanded test matrix to Ruby 3.0 ... 3.1 x Rails 6.1 ... 8.0 with blocking of incompatible combinations due to dependencies.
- Added a test matrix grid to the README
- Build issues were mostly around concurrent-ruby  needing to hold at 1.3.4 for Rails < 7.1
- `spec/support/migration_helper.rb` was causing failures due to breaking changes in Rails >= 7.0 . `ActiveRecord::MigrationContext.new()` no longer expected a method as the second argument but would accept a block. Removing the second argument of `schema_migration` fixed the issue.


joshbuker:
Not sure why GitHub actions isn't running on this PR. Going to force merge and see if it passes on main.

joshbuker:
Failing due to OAuth mismatch. Interesting...New version of OAuth dependency perhaps?

PR #379, opened 2025-06-16
Fix CI failures

### LINKED ISSUE #378: The test related to OAuth is failing
ref
- [Fix test suite failing (#377) · Sorcery/sorcery@cbfd127](https://github.com/Sorcery/sorcery/actions/runs/15655376104)
- [Sorcery Unmaintained - Looking for additional maintainers · Issue #350 · Sorcery/sorcery](https://github.com/Sorcery/sorcery/issues/350#issuecomment-2973057087)

willnet:
Fix #378

The test started to fail because the encoding of spaces was changed from "+" to "%20" by [this change](https://gitlab.com/oauth-xx/oauth2/-/merge_requests/633) included in oauth2 v2.0.10. After updating the expected values in the test to use "%20", the test passed.


willnet:
@joshbuker If you have time, I'd appreciate it if you could review this.

willnet:
It's a small change, so I was planning to merge it myself if I didn't hear back from anyone after a while. But I just noticed that the settings require one approval from a member to merge. So I'll wait for a review.

joshbuker (changes requested):
Changing the tests to make them pass is a bit of a red flag. Can you dig into why this change occurred before we update the test suite to match the results it's giving?

joshbuker (approved):
Ah, I missed your comment on it being from oauth 2.0.10.

joshbuker:
They say this is needed for Apple ID, but it's not clear if this is needed for other providers, or if it'll break other providers... Would be nice to test this in a staging environment to verify, but ultimately the oauth implementation we're using is horribly outdated and should be replaced with omniauth anyway. Can always roll it back if things break.

willnet:
I’ll go ahead and merge this PR for now.
If an issue is created later, I’ll handle it at that time.

PR #383, opened 2025-06-20
Drop support for versions below Ruby 3.2 and Rails 7.1

willnet:
Sorcery only supports the currently officially supported versions of Ruby and Rails.

Also remove conditionals for Rails versions older than 7.0, and mongoid versions older than 4.0

Mongoid 3.x is no longer relevant since its latest release (v3.1.7) depends on `activemodel ~> 3.2`, which is incompatible with current versions of Sorcery. See: https://github.com/mongodb/mongoid/blob/v3.1.7/mongoid.gemspec#L22

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -17,36 +17,14 @@ jobs:
       fail-fast: false
       matrix:
         ruby:
-          - '3.0'
-          - '3.1'
           - '3.2'
           - '3.3'
           - '3.4'
 
         rails:
-          - '61'
-          - '70'
           - '71'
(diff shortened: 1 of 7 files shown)


joshbuker:
@willnet , @Beartech needs to accept the org invite before you can request his review

joshbuker (approved):
Thanks!

joshbuker:
@willnet oh, can you remove the testing matrix from the readme? No longer needed as we removed the exclusions.

willnet:
@joshbuker Thank you for the feedback. I'll create a follow-up PR later.

PR #386, opened 2025-07-01
Limit the files included in the gem to only the necessary ones

willnet:
Previously, files like spec and .devcontainer were unnecessarily included in the gem. Update the settings so that only files under lib and documentation like README are included.


Beartech (approved):
Approved.

PR #390, opened 2025-08-20
Stop mocking models in controller specs and switch to using ActiveRecord directly

willnet:
Sorcery used to support multiple ORMs, so controller specs were heavily mocked to stay ORM-agnostic. Reference: [ORM-wide controller specs #586 by kirs · Pull Request #591 · NoamB/sorcery](https://github.com/NoamB/sorcery/pull/591#issuecomment-49626822)

However, since [v0.10.0](https://github.com/Sorcery/sorcery/blob/a5e2035d0a2c3cdabbeaf720949e9bf7d360986a/CHANGELOG.md#0100), non-ActiveRecord ORM adapters were moved out, and Sorcery now only supports ActiveRecord. In this situation, mocks hurt test readability and make maintenance harder, so I removed them and now use ActiveRecord directly in controller specs.

Also removed redundant logic and unnecessary hacks.

### DIFF
--- a/spec/rails_app/config/application.rb
+++ b/spec/rails_app/config/application.rb
@@ -11,7 +11,7 @@ require 'sorcery'
 
 module AppRoot
   class Application < Rails::Application
-    config.autoload_paths.reject! { |p| p =~ %r{//app/(\w+)$/} && !%w[controllers helpers mailers views].include?(Regexp.last_match(1)) }
+    config.autoload_paths.reject! { |p| p =~ %r{/app/(\w+)$} && !%w[controllers helpers mailers views].include?(Regexp.last_match(1)) }
     config.autoload_paths += ["#{config.root}/app/active_record"]
 
     # Settings in config/environments/* take precedence over those specified here.


joshbuker:
@brendon reached out separately offering to help with maintenance.

@brendon I'm adding you as a maintainer to the github org, if you'd like to assist @willnet with the PRs and getting things ready to cut a new release.

brendon:
Thanks @joshbuker, much appreciated. @willnet, I'm happy to take a look t this. I'll make a note to look through it tomorrow. We can also make a bit of a plan for what you want to get done going forward.

brendon (commented):
Looks pretty good. My only concern is that `expects` adds actual expectations to tests. Removal of these expectation clauses means we're not actually checking if a method is or isn't called on an object and how many times etc...

I think going forward this would have been a good PR to be split into maybe 3 smaller ones to make it easier to quickly see what's being changed :)

Great work though! Looking forward to seeing future progress on this gem! :D

joshbuker:
@brendon I got pretty far, but it's a complete rewrite from the ground up so drift is a non-issue. If you're wanting to improve Sorcery for long term use, I would recommend starting from it rather than starting over from scratch or trying to use the existing codebase.

joshbuker:
Any commits after 2023 (so 3694994 and later) should be reviewed for relevant changes to functionality. Where I left off was getting the current test suite 1:1 in the rework repo to ensure it at least had parity with what existed.

PR #393, opened 2025-09-29
Change test domains to use "example.com" wherever possible

willnet:
Previously used domains like "blabla.com", "bla.com", and "bar.com" could exist in reality. Since IANA reserves example.com for documentation and examples, I replaced them with example.com whenever possible.

The exception is test.host, which Rails uses as the default domain in tests (see: https://github.com/rails/rails/blob/d09c4bbfcadcd6b3af847a11608a63c0718158ed/actionpack/lib/action_dispatch/testing/test_request.rb#L11). Simply changing it to example.com causes tests to fail. While adding the following snippet to `spec_helper.rb` would make the tests pass, I decided to leave it as is since the motivation isn’t strong enough:

```ruby
config.before(:each, type: :controller) do
  request.host = 'example.com'
end
```

related url: https://github.com/Sorcery/sorcery/pull/387#discussion_r2264873785

### DIFF
--- a/lib/sorcery/test_helpers/internal.rb
+++ b/lib/sorcery/test_helpers/internal.rb
@@ -24,7 +24,7 @@ module Sorcery
       end
 
       def build_new_user(attributes_hash = nil)
-        user_attributes_hash = attributes_hash || { username: 'gizmo', email: 'bla@bla.com', password: 'secret' }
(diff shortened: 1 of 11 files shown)


brendon (approved):
Looks good!

PR #395, opened 2025-10-03
Set up the RuboCop environment

willnet:
RuboCop has been partially introduced so far, but I want to run it properly from now on. I installed the latest rubocop, rubocop-rspec, and rubocop-performance and wired them into CI. To keep CI behavior consistent, each gem’s version is pinned except for patch versions.

All cops that currently fail are listed in .rubocop_todo.yml. I plan to work down this todo over time, but I do not intend to aggressively fix cops myself. By leaving RuboCop todos as good first issues, I want to create an entry point for new contributors.

I don’t necessarily want to fix every cop in the todo. We’ll weigh the cost of following a cop’s rule against its benefit, and disable the ones where the cost outweighs the value.


brendon (approved):
Great work! :)

PR #400, opened 2025-10-31
Add a dependency on Rails 7.1 or higher

willnet:
In https://github.com/Sorcery/sorcery/pull/383￼, we dropped support for Rails versions below 7.1, but there was no way to enforce this requirement for Sorcery users. For example, it was still possible to use the latest version of Sorcery with Rails 6. To prevent issues caused by accidentally upgrading Sorcery while staying on an older version of Rails, we explicitly added a dependency on railties.

We depend on railties instead of rails to reduce the number of dependencies. The rails gem depends on all other Rails components like activerecord, but some developers intentionally avoid unused components in their applications. This change keeps the dependency minimal for those users.


brendon (approved):

brendon:
Excellent! :D

PR #405, opened 2025-11-28
Remove Rails-version-based branchings that are no longer needed

willnet:
Sorcery now supports Rails 7.1 and above, so conditional checks for Rails 7 or earlier are unnecessary.


atolix (approved):
approved.

PR #406, opened 2025-11-28
Make redirect_to_before_login_path capable of redirecting to other hosts

willnet:
Related PR: #351

Add an `allow_other_host` argument to the `redirect_to_before_login_path` method so it can redirect to other hosts.


brendon (approved):
Looks great! :)

PR #413, opened 2025-12-06
Add instructions on how to release a gem in MAINTAINING.md

willnet:
I released a new gem myself and confirmed that `bundle exec rake release` works, so I added that to MAINTAINING.md.
I’m not sure what exactly should go in [this TODO section](https://github.com/Sorcery/sorcery/blob/1baeb219e7e15ad3b902b0e9dbdef48eaa616df5/MAINTAINING.md#L15-L17), so for now I only wrote what I could.

### DIFF
--- a/MAINTAINING.md
+++ b/MAINTAINING.md
@@ -52,10 +52,9 @@ NOTE: `X.Y.Z` and `vX.Y.Z` are given as examples, and should be replaced with
 1. Stage your changes and create a commit
    1. `git add -A`
    1. `git commit -m "Release vX.Y.Z"`
-1. TODO: Gem Release (WIP)
+1. Gem Release
    1. `cd <dir>`
-   1. `gem build`
-   1. `gem push <filename>`
+   1. `bundle exec rake release`
 1. Create GitHub Release
    1. Create a new release via GitHub interface at https://github.com/Sorcery/sorcery/releases/new
    1. Use tag `vX.Y.Z` and title `vX.Y.Z`


willnet:
@joshbuker @brendon I tried introducing [rubygems/release-gem](https://github.com/rubygems/release-gem) in a few gems that I maintain, and it’s really convenient. If you make a workflow like [ci_logger/.github/workflows/push_gem.yml at main · willnet/ci_logger](https://github.com/willnet/ci_logger/blob/b8aa6a622a332a5917fcaaee34a4c02dcbdae7ad/.github/workflows/push_gem.yml), you can release to rubygems.org and create a GitHub Release at the same time through it. I’d like to apply this setup to Sorcery as well.

However, if we use [this workflow](https://github.com/willnet/ci_logger/blob/b8aa6a622a332a5917fcaaee34a4c02dcbdae7ad/.github/workflows/push_gem.yml) as-is, anyone who can run the workflow will also be able to release. So compared to now, the number of people who can release will probably increase. And if more people can release, the takeover risk goes up as well.

As a way to reduce the risk, we can use GitHub’s built-in feature called [environment](https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/manage-environments). By creating an environment, we can add restrictions like requiring approval before the workflow can proceed.

In my personal opinion, it would be good to switch to releasing via release-gem, and require one approval from someone other than the person who started the workflow for releases. What do you think? I’d like to hear your opinions.

PR #419, opened 2025-12-13
Fix rubocop rspec offenses

beengine:

- [X] Description of changes
- [X] Changes have related RSpec tests that ensure functionality does not break

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -62,3 +62,11 @@ Style/Documentation:
 Style/OptionalBooleanParameter:
   Exclude:
     - 'lib/**/*'
+
+####################
+## RSpec Style    ##
+####################
+
+# Allow up to 5 expectations per test for integration tests
+RSpec/MultipleExpectations:
+  Max: 5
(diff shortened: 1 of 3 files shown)


spec/controllers/controller_oauth2_spec.rb
```diff
@@ -379,11 +379,11 @@
   end
 
   def stub_all_oauth2_requests!
-    access_token = double(OAuth2::AccessToken)
+    access_token = double(OAuth2::AccessToken) # rubocop:disable RSpec/VerifiedDoubles
```
willnet: Why is this disabled?

willnet: Thank you for the explanation. After checking, it seems that `OAuth2::AccessToken#token_param=` was already removed back in 2011.
All places that use this method will need to be fixed, but it would be better to handle that separately, so this PR is fine as it is.

ref: [refactor of token and strategy · ruby-oauth/oauth2@0b0c322](https://github.com/ruby-oauth/oauth2/commit/0b0c322dceadaf00227431bc8160fb2bf1f26837)

willnet (approved):

PR #420, opened 2025-12-23
Remove OAuth2 authentication using LiveID

willnet:
OAuth that uses LiveID is no longer available. As a result, this feature cannot be used by anyone at the moment.

ref: [Migrating from Live SDK to Microsoft Graph - OneDrive dev center | Microsoft Learn](https://learn.microsoft.com/en-us/onedrive/developer/rest-api/concepts/migrating-from-live-sdk?view=odsp-graph-online)


brendon (approved):
Looks good to me! :)

PR #423, opened 2025-12-27
Inline shared_examples that are only referenced in one place

willnet:
Previously, Sorcery supported multiple ORMs, and shared_examples were used to run the same tests across different ORMs as shown below:

activerecord: https://github.com/Sorcery/sorcery/blob/a69742e0f5b0e18a76367e0ead747272c058b1ec/spec/active_record/user_activation_spec.rb#L16

datamapper: https://github.com/Sorcery/sorcery/blob/a69742e0f5b0e18a76367e0ead747272c058b1ec/spec/datamapper/user_activation_spec.rb#L8

However, since only ActiveRecord is currently being tested, there is little value in keeping these shared_examples. Therefore, I am removing the shared_examples and inlining the code instead.


brendon (approved):
Great work :)

PR #427, opened 2026-01-17
Ensure `Timecop.return` always runs after tests

willnet:
Some tests use `Timecop.freeze` but do not call `Timecop.return` afterward (e.g. the test in https://github.com/Sorcery/sorcery/blob/4e77ab2e379fd5d938f42a59b17696ad048fed25/spec/controllers/controller_oauth2_spec.rb#L324-L331 ).

This can cause flaky tests, so always call `Timecop.return` in the teardown for all tests.

### DIFF
--- a/spec/active_record/user_activation_spec.rb
+++ b/spec/active_record/user_activation_spec.rb
@@ -297,9 +297,9 @@ describe User, :active_record do
         sorcery_model_property_set(:activation_token_expiration_period, 0.1)
         user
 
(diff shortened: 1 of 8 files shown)


brendon (changes requested):
I wonder if it'd be cleaner to use the block form:

```
Timecop.freeze(Date.today + 30) do
  assert joe.mortgage_due?
end
```

wherever freeze is used. That way the effects are isolated to the block and we don't need to the global unfreeze?

PR #431, opened 2026-02-21
Add tests for behavior when a CSRF verification error occurs, which did not previously exist.

willnet:
Add tests for the code at [https://github.com/Sorcery/sorcery/blob/883852e59095333d63c9fe4d3517011be497e676/lib/sorcery/controller.rb#L139-L144](https://github.com/Sorcery/sorcery/blob/883852e59095333d63c9fe4d3517011be497e676/lib/sorcery/controller.rb#L139-L144) because there were no corresponding tests.

### DIFF
--- a/spec/controllers/sorcery_controller_spec.rb
+++ b/spec/controllers/sorcery_controller_spec.rb
@@ -175,6 +175,7 @@ describe SorceryController, type: :controller do
       it 'clears the remember_me_token cookie when CSRF token is invalid' do
         session[:user_id] = user.id.to_s
         request.cookies[:remember_me_token] = 'test_token'
+        expect(request.cookies[:remember_me_token]).to eq 'test_token'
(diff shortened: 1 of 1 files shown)


lib/sorcery/controller.rb
```diff
@@ -140,7 +140,7 @@ def auto_login(user, _should_remember = false)
       def handle_unverified_request
         cookies[:remember_me_token] = nil
         @current_user = nil
-        super # call the default behaviour which resets the session
+        super # call the default behaviour
```
willnet: This does not always reset the session

ref: https://api.rubyonrails.org/classes/ActionController/RequestForgeryProtection/ClassMethods.html#method-i-protect_from_forgery

brendon (changes requested):

brendon (approved):
Looks good :)

PR #438, opened 2026-05-12
Expand controller coverage for Phase 1.3

willnet:
Cover Sorcery::Controller core and controller submodule edge cases so the refactor roadmap has stronger regression protection. Mark Phase 1.3 complete after verifying the full spec suite passes.

### DIFF
--- a/spec/controllers/sorcery_controller_spec.rb
+++ b/spec/controllers/sorcery_controller_spec.rb
@@ -87,6 +87,18 @@ describe SorceryController, type: :controller do
           expect(session[:user_id]).to be_nil
         end
       end
+
+      it 'yields the failure reason when authentication fails with a reason' do
+        yielded = nil
+        allow(User).to receive(:authenticate).with('bla@example.com', 'wrong').and_yield(nil, :invalid_password)
+
+        result = controller.login('bla@example.com', 'wrong') do |user, failure_reason|
+          yielded = [user, failure_reason]
+        end
+
+        expect(result).to be_nil
+        expect(yielded).to eq([nil, :invalid_password])
+      end
     end
 
     describe '#login!' do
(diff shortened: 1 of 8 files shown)


spec/controllers/sorcery_controller_spec.rb
```diff
@@ -87,6 +87,18 @@
           expect(session[:user_id]).to be_nil
         end
       end
+
+      it 'yields the failure reason when authentication fails with a reason' do
+        yielded = nil
+        allow(User).to receive(:authenticate).with('bla@example.com', 'wrong').and_yield(nil, :invalid_password)
```
brendon, line 93: This test isn't really testing the code since the entire interaction is mocked.

brendon (changes requested):
Good work. I think get the AI to generate a new code coverage report to check that it's improved to 100% coverage in that area. Also, given the mock issues I found, it might pay to double-check the other tests aren't doing similar things.

spec/controllers/sorcery_controller_spec.rb
willnet: Other parts of this test file also use a lot of mocks. I guess it's because the ORM is abstracted by an adapter, and they wanted to test the controller without touching the database.

But as you said, relying too much on mocks means the code isn't being fully tested. So, along with adding coverage, I'll create a commit to reduce the overall use of mocks and fix the tests.

brendon: Sounds like a good plan :D

willnet: If I simply add the commit to this branch, it will conflict with #440. So I plan to add the commit after #440 is merged.

brendon: Roger :)

brendon (approved):

